PR #75, opened 2021-11-04
MSK documentation

amboulouma:
MSK documentation with create and describe cluster examples

### DIFF
--- a/content/en/aws/managed-streaming-for-kafka/index.md
+++ b/content/en/aws/managed-streaming-for-kafka/index.md
@@ -6,9 +6,9 @@ description: >
   Managed Streaming for Kafka (MSK)
 ---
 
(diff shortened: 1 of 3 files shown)


whummer:
Great start, thanks @amboulouma ! 👍 Added a few minor comments/nitpicks..

Would be nice if we also added a small sample of how to use Lambda event source mappings from MSK topics.. Thanks

whummer:
Thanks for updating the PR @amboulouma . Would be great if we could further extend the docs with these two aspects:

* short example on how to publish/subscribe to Kafka messages after the cluster has been created (e.g., via the [`kafka-console-producer`/`kafka-console-consumer`](https://kafka.apache.org/quickstart) command lines that ship with Kafka)
* small sample of how to use Lambda event source mappings from MSK topics (triggering Lambda functions from messages sent to the Kafka topics).. see here: https://docs.aws.amazon.com/lambda/latest/dg/with-msk.html#services-msk-topic-add 

Thanks!

PR #76, opened 2021-11-05
change invalid terraform command deploy to apply

captainmoha:
Terraform does not have a deploy command, so it will not work. It has apply command which provisions the resources specified in the config file. Changing this to apply make this work.


whummer:
Great, thanks for fixing this @captainmoha ! 👍

PR #84, opened 2021-12-09
Add docs for stay open mode

dfangl:
This PR adds docs and information about configuration variables introduced with https://github.com/localstack/localstack/pull/5088.

It also explains the difference between the `exec` docker reuse mode and the stay-open mode and mentiones limitations of the latter.

### DIFF
--- a/content/en/localstack/lambda-executors.md
+++ b/content/en/localstack/lambda-executors.md
@@ -45,10 +45,29 @@ A typical invocation of a dummy python lambda can take around 3 seconds from sta
 All supported lambda types can be used with this executor.
 
 
-### Docker re-use
+### Docker reuse
 
 Configuration: `LAMBDA_EXECUTOR=docker-reuse`
+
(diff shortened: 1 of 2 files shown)


whummer (approved):
Great update to the docs @dfangl ! 👍

content/en/localstack/lambda-executors.md
```diff
@@ -45,10 +45,29 @@ A typical invocation of a dummy python lambda can take around 3 seconds from sta
 All supported lambda types can be used with this executor.
 
 
-### Docker re-use
+### Docker reuse
 
 Configuration: `LAMBDA_EXECUTOR=docker-reuse`
 
+#### Stay-open mode
+LocalStack allows to use the stay-open mode of its lambda containers.
```
whummer, line 53: nit: We're usually writing `Lambda` with upper-case L, to make it clear that we're referring to the service name (not critical at all, though..).

PR #91, opened 2021-12-20
Add hint to get-started about LAMBDA_DOCKER_NETWORK

dfangl:
Currently, it seems not clear to users when to set this.
Until we can get rid of the necessity of this variable, it should be mentioned more prominently.

Addresses https://github.com/localstack/localstack/issues/5130 https://github.com/localstack/localstack/issues/5140


whummer:
Thanks @dfangl 👍

PR #92, opened 2021-12-21
CI: Add broken link checker for docs

HarshCasper:
## Description

This PR:

- Adds a GHA to look for broken links
- Fix noticeable broken links

### DIFF
--- a/.github/workflows/markdown.links.config.json
+++ b/.github/workflows/markdown.links.config.json
@@ -1,7 +1,19 @@
 {
     "ignorePatterns": [
 		{
-			"pattern": ""
-		}
+			"pattern": "https://kinesis.us-east-1.amazonaws.com/"
+		},
+        {
+            "pattern": "http://(.*)localhost(.*)"
+        },
+        {
+            "pattern": "https://mvnrepository.com/artifact"
+        },
+        {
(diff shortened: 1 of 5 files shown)


.github/workflows/markdown.links.config.json
```diff
@@ -0,0 +1,25 @@
...
+		},
+        {
+            "pattern": "http://localhost:4566"
+        },
+        {
+            "pattern": "http://localhost:4598"
+        },
+        {
+            "pattern": "http://my-domain.us-east-1.es.localhost.localstack.cloud:4566"
+        },
```
thrau, line 14: according to [the docs](https://github.com/tcort/markdown-link-check#markdownlinkcheckmarkdown-opts-callback), the patterns here can be regular expressions. can we maybe generalize the localhost pattern into a regex? not sure how to express it correctly here, but something like `http://(.*).localhost.localstack.cloud(.*)`, ... or really anything with localhost in it ;-)

thrau (changes requested):
this is very cool and will be super useful! i was hoping we could maybe just generalize the ignore patterns using regex before merging.

thrau (approved):
thanks for this!

Issue #70, opened 2021-10-26
Docs enhancement: Adding a copy button in the command sections


aminblm:
Adding a copy button in the command sections would allow faster copy-pasta.

ankitagit11:
I want to work on this issue. Please assign it to me.

lukqw:
@ankitagit11 are you still interested in this issue?

lukqw:
closing since this was already implemented in https://github.com/localstack/docs/pull/719

PR #94, opened 2021-12-26
feat: Enable Prism for copy code button

iamrajiv:
## Description

Fixes: #70

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -142,7 +142,7 @@ algolia_docsearch = false
 offlineSearch = true
 
 # Enable syntax highlighting and copy buttons on code blocks with Prism
-prism_syntax_highlighting = false
+prism_syntax_highlighting = true
 
 # User interface configuration
 [params.ui]


iamrajiv:
@thrau @whummer please take a look.

alexrashed:
Unfortunately the prism syntax highlighting is _not_ compatible with [our `command` shortcode](https://github.com/localstack/docs/tree/main/layouts/shortcodes) (i.e. it breaks when you activate prism). This shortcode makes sure that you can easily copy our commands in the docs (where is prevents the selection of the `$ ` prefix).
You can verify that by starting the hugo page (`hugo serve`) and then trying to select or copy commands ([for example the pip install command here](https://docs.localstack.cloud/get-started/#installation)).
I'm afraid that you will need to (heavily?) adjust the shortcode before you can enable the prism syntax highlighting.

config.toml
```diff
@@ -142,7 +142,7 @@ algolia_docsearch = false
 offlineSearch = true
 
 # Enable syntax highlighting and copy buttons on code blocks with Prism
-prism_syntax_highlighting = false
+prism_syntax_highlighting = true
```
alexrashed, line 145: Unfortunately this change needs to come with a refactoring of the `command` shortcode :(

alexrashed (changes requested):

config.toml
iamrajiv: Yeah sure @alexrashed will try to edit command shortcode and will debug this.

PR #107, opened 2022-01-24
Fix `INIT_SCRIPTS_PATH` default location

56KBs:
The default location for `INIT_SCRIPTS_PATH` incorrectly contains a trailing dot.


thrau:
great catch, thanks!

Issue #87, opened 2021-12-14
Add Open Graph image


iamrajiv:
## Description

The Open Graph images are not generated when we share https://docs.localstack.cloud/ over social media.

![Screenshot 2021-12-14 at 9 14 39 AM](https://user-images.githubusercontent.com/42106787/145929386-e8d5f107-04e6-43f0-b97b-0c539920a316.jpg)

## References

- Use https://developers.facebook.com/tools/debug/ to validate results for Facebook.
- Use https://www.linkedin.com/post-inspector/ to validate results for LinkedIn.
- Use https://cards-dev.twitter.com/validator to validate results for Twitter.

iamrajiv:
@whummer @thrau

iamrajiv:
@thrau @whummer

HarshCasper:
Hi @iamrajiv 👋 

Would you like to work on it? I have assigned you the issue, but feel free to issue if there is something else on priority 😄

Bisht13:
Taking this up ✨

PR #115, opened 2022-02-19
feat: Add Open Graph image

iamrajiv:
Fixes: #87

Screenshot 2022-02-20 at 4 58 12 AM

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -93,6 +93,7 @@ weight = 1
 section = ["HTML", "RSS"]
 
 [params]
+images = ['https://localstack.cloud/images/localstack.png']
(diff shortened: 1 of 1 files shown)


thrau (commented):
this is a great idea! i don't think the preview looks very good. could we try with the social media teaser?

localstack-readme-header

iamrajiv:
@thrau Yeah the quality of the image is not good. What is a social media teaser can you provide the image link? Also, I would suggest we should have a separate branding repository that contains all branding assets of LocalStack. We should also add the Branding option on the https://localstack.cloud/ website. I am happy to work on this.

PR #119, opened 2022-03-04
Update pulumi service endpoints and reorder config

dominikschubert:
updates the endpoints to all that are currently available in pulumi


whummer (approved):
LGTM 👍

PR #124, opened 2022-03-11
updated terraform installation

macnev2013:
# Added
- Reference installation links

# Updated
- Terraform installation steps
Note: Installation steps taken from [here](https://learn.hashicorp.com/tutorials/terraform/install-cli)

### DIFF
--- a/content/en/developer-guide/setup.md
+++ b/content/en/developer-guide/setup.md
@@ -69,12 +69,14 @@ Below are some basic installation instructions for the dependencies you will nee
   ```
 * Terraform
   ```bash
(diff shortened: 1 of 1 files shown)


thrau (changes requested):
thanks for improving the docs @macnev2013!
i think this is very useful for debian users, but the previous installation instruction was cross-platform and would work on different distros and mac.
maybe we can separate the two installation instructions?

we have a cool feature for this sort of thing, the `tabpane` shortcode. have a look here: https://docs.localstack.cloud/integrations/spring-cloud-function/#setting-up-deployment
this is what the code looks like: https://github.com/localstack/docs/blob/e7c344addc0dcb88eb13d850755c198572814cb0/content/en/integrations/spring-cloud-function/index.md?plain=1#L595-L596

macnev2013:
Links to the installations are already added. This is not applicable anymore.

PR #126, opened 2022-03-15
CI: Enable PR previews for docs

HarshCasper:
## Description

This PR:

- Adds a custom workflow to deploy docs using `npx` and `surge`
- Adds an action to install, build and store the build as an artifact
- Adds a workflow step to put up a GitHub comment on respective PRs after deploying


thrau (approved):
:100:

PR #129, opened 2022-03-16
add bind mount example

calvernaz:
highlights the need to bind mount the docker socket when running localstack container and lambda executors in docker mode

Practising learnings from today's workshop.


alexrashed (approved):
🚀

dfangl (approved):

PR #131, opened 2022-03-25
development environment setup instructions

ksh7:
These instructions were removed from the localstack `README` file.


FelipeRuizGarcia:
@ksh7 what do you thing about create a new guide to describe the next steps to run localstack, after the setup? Like #137

thrau:
thanks for adding this @ksh7, and sorry for taking so long! I also like the idea of @FelipeRuizGarcia, however since this is developer docs, it's probably something the core devs should come up with :-)

PR #138, opened 2022-04-13
Add tutorial on ECR EKS integration

dfangl:
Due to recent interest, I created a small sample to demonstrate the integration between EKS and ECR and its pitfalls (mainly the possibility of failure to resolve `localhost.localstack.cloud`).

It is a small example, with the necessary explanation at the beginning, with a small sample to demonstrate (nothing special, just creating a pod up to the point where the image is pulled) afterwards, using awscli. With the correct configuration of `EXTERNAL_HOSTNAME`, any IaC tool should work as well though.

### DIFF
--- a/content/en/aws/elastic-kubernetes-service/index.md
+++ b/content/en/aws/elastic-kubernetes-service/index.md
@@ -52,6 +52,148 @@ NAME                 TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
 service/kubernetes   ClusterIP   10.43.0.1    <none>        443/TCP   70s
 {{</ command >}}
 
+### Use images pushed to ECR in EKS
+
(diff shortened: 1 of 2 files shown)


dfangl:
Error in linting seems to be caused by https://github.com/gaurav-nelson/github-action-markdown-link-check/issues/132 which was caused by https://github.blog/2022-04-12-git-security-vulnerability-announced/ .

whummer (approved):
Awesome, thanks for adding this sample @dfangl ! 💯 

Linter check is still failing - is that something we can look into in a follow-up PR @HarshCasper ?

HarshCasper:
> Awesome, thanks for adding this sample @dfangl ! 💯
> 
> Linter check is still failing - is that something we can look into in a follow-up PR @HarshCasper ?

I will fix that or migrate it to an alternative if this persists.

Issue #98, opened 2022-01-02
Docs for localstack profiles


thrau:
There's currently no documentation for the `~/.localstack` environment profiles as described here: https://github.com/localstack/localstack/pull/5022

PR #140, opened 2022-04-15
DOCS: Add docs about configuration profiles in LocalStack

HarshCasper:
Fixes https://github.com/localstack/docs/issues/98

Preview doc: https://localstack-docs-preview-pr-140.surge.sh/localstack/configuration/#profiles

### DIFF
--- a/content/en/localstack/configuration.md
+++ b/content/en/localstack/configuration.md
@@ -213,7 +213,7 @@ Some of the services can be configured to switch to a particular provider:
 
 ## Profiles
 
-LocalStack support configuration profiles which is stored in the `~/.localstack` config directory. A configuration profile is a set of environment variables stored in an `.env` file in the LocalStack config directory. Here is an example of what configuration profiles might look like:
+LocalStack supports configuration profiles which are stored in the `~/.localstack` config directory. A configuration profile is a set of environment variables stored in an `.env` file in the LocalStack config directory. Here is an example of what configuration profiles might look like:
 
 ```sh
 % tree ~/.localstack 
@@ -223,10 +223,11 @@ LocalStack support configuration profiles which is stored in the `~/.localstack`
 └── pro.env
 ```
 
-Let's say you wish to have a look at the `dev.env` profile. You can do so by pushing the following command:
+Here is an example of what a specific environment profile looks like
 
 ```sh
-% cat ~/.localstack/dev.env
+% cat ~/.localstack/pro-debug.env
+LOCALSTACK_API_KEY=XXXXX
 DEBUG=1
 DEVELOP=1
 ```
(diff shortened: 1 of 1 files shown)


content/en/localstack/configuration.md
```diff
@@ -168,6 +168,37 @@ Some of the services can be configured to switch to a particular provider:
 | `KMS_PROVIDER` |  `moto` (default) and `local-kms` |
 | `SQS_PROVIDER` |  `moto` (default) and `elasticmq` |
 
+## Profiles
+
+LocalStack support configuration profiles which is stored in the `~/.localstack` config directory. A configuration profile is a set of environment variables stored in an `.env` file in the LocalStack config directory. Here is an example of what configuration profiles might look like:
```
thrau, line 173: typo: LocalStack supports ..., which are ...

content/en/localstack/configuration.md
```diff
@@ -168,6 +168,37 @@ Some of the services can be configured to switch to a particular provider:
...
+
+```sh
+% tree ~/.localstack 
+/home/username/.localstack
+├── default.env
+├── dev.env
+└── pro.env
+```
+
+Let's say you wish to have a look at the `dev.env` profile. You can do so by pushing the following command:
```
thrau, line 183: If someone created an `.env` file I think we can assume that they know how to display it. I think we could reformulate this into "here is an example of what a specific environment profile looks like"

content/en/localstack/configuration.md
```diff
@@ -168,6 +168,37 @@ Some of the services can be configured to switch to a particular provider:
...
+
+```sh
+% cat ~/.localstack/dev.env
+DEBUG=1
+DEVELOP=1
```
thrau, line 188: maybe we can call it `pro-debug.env`, and add the `LOCALSTACK_API_KEY` with a dummy value? i think that will be what people use most often :-)

content/en/localstack/configuration.md
```diff
@@ -168,6 +168,37 @@ Some of the services can be configured to switch to a particular provider:
...
+```
+
+You can load a profile by either setting the `env` variable `CONFIG_PROFILE=<profile>` or the `--profile=<profile>` CLI flag when using the CLI. Let's take an example to load the `dev.env` profile file if it exists:
+
+{{< command >}}
+python -m localstack.cli.main --profile=dev start --host
+{{< / command >}}
+
+If no profile is specified, the `default.env` profile will be loaded. While explicitly specified, the environment variables will always overwrite the profile. To display the config environment variables, you can use the following command:
```
thrau, line 197: just for future docs: having each sentence in a separate line helps greatly for viewing git diffs for later changes

thrau (changes requested):
Thanks for adding this @HarshCasper! Just a few comments to address. Maybe we can also add a reference to the CLI getting started page.

Issue #147, opened 2022-04-26
Error building site: EXECUTE-AS-TEMPLATE | Fedora


FelipeRuizGarcia:
Unable to run Hugo server in the main branch 
```commit 287fa6f102652328c8fc54f8e0444879e105f150 (HEAD -> main, origin/main, origin/HEAD) ```

After clone and submodules updated, output says:
```
`~/repos/docs │ main  hugo serve                                                                                                    ✔ │ 8m 17s │ 08:53:18 PM 
Start building sites … 

Built in 1286 ms
Error: Error building site: EXECUTE-AS-TEMPLATE: failed to transform "json/offline-search-index.json" (application/json): failed to parse Resource "json/offline-search-index.json" as Template:: template: json/offline-search-index.json:5: unclosed action
 ~/repos/docs │ main 
 `
```

SYSTEM
FEDORA 35

Hugo Static Site Generator v0.80.0/extended linux/amd64 BuildDate: unknown

FelipeRuizGarcia:
Task: test with the recent Hugo version

FelipeRuizGarcia:
Problem Solved:
The root cause was the Hugo Version: The fix is included in the latest Hugo version  `v0.97.3`

Solution: The easiest way is get the latest hugo from brew! 
https://formulae.brew.sh/formula/hugo#default

PR #148, opened 2022-04-27
Suggest use the updated version of Hugo

FelipeRuizGarcia:
There are some circular issues on Hugo 0.8X and 0.9.X, more info at https://github.com/gohugoio/hugo/releases/tag/v0.97.3.

and
https://github.com/localstack/docs/issues/147

We can simply suggest to use the latest v0.97.3 to avoid well-know issues


FelipeRuizGarcia:
@HarshCasper @whummer

PR #150, opened 2022-05-03
Document DISABLE_TERM_HANDLER, update docker run command in getting-started

dfangl:
Port 4571 is not used by any default configuration (only legacy elasticsearch), ports 4510-4559 are however used more frequently.

I also removed the hint to the "SERVICES" variable, as we see a lot of misuse in support. Up to discussion!

Main target of this PR is the addition of documentation regarding the environment variable introduced in https://github.com/localstack/localstack/pull/6000 .

This variable disables "orderly" shutdown of LocalStack by preventing passing signals from docker-entrypoint.sh to supervisor.


thrau (approved):
:100:

PR #157, opened 2022-05-17
fix capitalization of LocalStack in affected files

lukqw:
Noticed while reading the docs that the capitalization of LocalStack is inconsistent.
This PR corrects these mistakes.

### DIFF
--- a/content/en/aws/cognito/index.md
+++ b/content/en/aws/cognito/index.md
@@ -167,7 +167,7 @@ Note that the value of the `redirect_uri` parameter must match the value provide
 
 ## Serverless and Cognito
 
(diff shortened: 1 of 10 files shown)


silv-io (approved):
LGTM! 🚀 

Found another occurrence in `content/en/integrations/serverless-framework/index.md:L151` with `ripgrep`

lukqw:
Thanks for double checking! I found that occurence as well. As it turns out, this occurence correlates to log output, which can be found [here](https://docs.localstack.cloud/integrations/serverless-framework/#deploying-to-localstack). Seems like the error is produced by our loggers. 👀

PR #165, opened 2022-05-23
Add RDSnotes about db-name postgres

MarcelStranak:


steffyP (approved):
Thanks for adding this info, @MarcelStranak! 👍

PR #177, opened 2022-06-02
docs: add drone ci to list of supported services

HarshCasper:


content/en/ci/drone-ci/index.md
```diff
@@ -0,0 +1,52 @@
...
+  Use LocalStack in [Drone CI](https://drone.io/)
+---
+
+This guide shows you how to start LocalStack in a Drone CI pipeline.
+
+## Drone CI Pipelines
+
+There are a few restrictions in Drone CI Pipelines that make it hard to customize the behavior of Docker. For example, mounting the host machine Docker socket is considered highly insecure, and hence alleviated privileges are required to run commands like `localstack wait`. Learn more about configuring Docker for Drone CI pipelines over [their official documentation](https://docs.drone.io/pipeline/docker/overview/).
+
+A possible work-around to mount the Docker socket (`/var/run/docker.sock`) into the container to communicate with the docker daemon on the host is by exposing the port `2375`, which is reachable in the main LocalStack container. However, it requires changes to the typical Docker-Compose setup.
```
whummer, line 15:
```suggestion
A possible alternative to mounting the Docker socket (`/var/run/docker.sock`) into the container to communicate with the Docker daemon on the host is to expose port `2375` and make it reachable in the main LocalStack container. However, it requires changes to the typical Docker-Compose setup.
```

whummer: Not sure if my suggestion matches what we're intending to express here - feel free to adjust as you see fit! 👍 

Could we perhaps add a small config snippet that illustrates how this can be achieved, to make the Docker daemon available? 

Otherwise, if we cannot easily get the LocalStack container to talk to the Docker engine (and spin up new containers), it would be great if we could document the limitations (e.g., that only the `local` Lambda executor will work in that case, i.e., users should configure `LAMBDA_EXECUTOR=local`).

whummer (approved):
Nice job @HarshCasper - great to see that we have another CI integration that we can easily provide for our users! 🎉 Only a few minor nitpick comments..

thrau (approved):
LGTM (+1 on waldemar's comments)

jimsheldon:
Looking forward to this getting merged!

PR #178, opened 2022-06-05
update requirements python version

yktakaha4:
https://github.com/localstack/localstack/pull/6050 has been merged, it appears that Python 3.10 or higher is now required.

I tried to `make install` with Python 3.9, and the following notation gave me a TypeError.
https://github.com/localstack/localstack/blob/master/localstack/testing/aws/cloudformation_utils.py#L9

ref: https://peps.python.org/pep-0604/


whummer (approved):
Good catch, thanks for fixing this @yktakaha4 !

PR #188, opened 2022-06-29
New filesystem layout

viren-nadkarni:
- Adds new page describing the new filesystem hierarchy
- Updates references to deprecated vars `DATA_DIR` and `HOST_TMP_DIR` everywhere

Followup to https://github.com/localstack/localstack/pull/6302/

PR #190, opened 2022-06-29
Drop mentions to persistence in Community

giograno:
In the docs we still refer to the old persistence mechanism, based on the record-and-replay approach, which is currently not supported anymore. 
This PR starts updating that part of the documentation.

### DIFF
--- a/content/en/localstack/persistence-mechanism/index.md
+++ b/content/en/localstack/persistence-mechanism/index.md
@@ -11,7 +11,8 @@ Commonly, you may simply have a local development server that relies on a non-ep
 
 While the persistence mechanism covers most services, not all of them are supported yet.
 Please make sure to check the [feature coverage page]({{< ref "feature-coverage" >}}) to see whether your desired services are covered.
-Please note that the coverage is only guaranteed for the _Pro_ version, while the _Community_ version attempts to restore the state on a best-effort basis using a *record-and-replay* approach (more on that in the [Technical Details]({{< ref "#technical-details" >}}) section).
+While in the past we supported a version of persistence -- available in the _Community_ version -- based on a *record-and-replay* approach (basically, storing API calls and re-running them on restart), we discontinued such a feature since [0.13.1](https://github.com/localstack/localstack/releases/tag/v0.13.1).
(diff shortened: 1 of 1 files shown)


content/en/localstack/persistence-mechanism/index.md
```diff
@@ -48,46 +49,21 @@ Otherwise, the endpoint will inform you that the mechanism is disabled.
...
 ![Structure of the DATA_DIR](datadir_structure.png)
 
-This approach does not suffer from the same limitations as *record-and-replay*.
 Restoring the state -- even for large projects -- usually only takes a few milliseconds.
-Moreover, since the files store accurate snapshots of application state it can restore a state that is identical to the one prior to the reboot.
+Moreover, since the files store accurate snapshots of the application state they can restore a state that is identical to the before the reboot.
 
-However, the limitation of this approach is that it is not quite as generic.
-The *record-and-replay* approach can treat the AWS services as "black-boxes", i.e. the persistence mechanism is entirely decoupled from the underlying services, while the *serialized state* approach is coupled to the serialization mechanism of each service.
+The most notable limitation of this approach it's its generalizability.
```
whummer, line 67:
```suggestion
The most notable limitation of this approach is its generalizability.
```

content/en/localstack/persistence-mechanism/index.md
```diff
@@ -48,46 +49,21 @@ Otherwise, the endpoint will inform you that the mechanism is disabled.
...
+In fact, despite most of the supported AWS services serialize their state in a similar way, particular services might implement such a serialization in a completely different fashion.
```
whummer, line 68: Wondering if we could simplify the last paragraph (last 3 lines) here a bit. I think users probably care most about which services have support for persistence (which should be almost all services at this stage..?), not so much about the limitations of the internal storage mechanism, or how it differs between services. 

We had this paragraph in here previously to compare the 2 persistence approaches, but probably we can simply remove lines 67-69? (and instead mention any services that do **not** yet support persistence, if there are any..)

whummer (approved):
Thanks for updating this @giograno ! Just minor nitpicks - and one comment about a paragraph that could be removed..

viren-nadkarni:
There is some overlap here with https://github.com/localstack/docs/pull/188. Also we should target the `v1` branch, not `main`.

PR #193, opened 2022-07-05
fix opensearch custom cluster sample

dfangl:
While testing for PR #192 , I noticed some inconsistencies in the sample, which are fixed by this PR:
* Domain name in command did not match output or subsequent operations
* Opensearch domain contained `.es.` instead of `.opensearch.`, leading to it being routed to S3.


alexrashed (approved):
Great catch! Thanks for fixing it right away! 🔝

PR #197, opened 2022-07-12
Update docker port range on pro documentation

darwin403:
The docker port range doesn't seem to reflect the latest port ranges. I have updated it.


dfangl (approved):
Absolutely correct, thanks for the contribution!

PR #200, opened 2022-07-12
Update implemented CloudFormation models

dominikschubert:


whummer (approved):
Looks awesome @dominikschubert  - very valuable info for our users, as well as super helpful for us internally!! 💯

PR #203, opened 2022-07-13
minor: rename Synopsys->Synopsis

whummer:
minor nit: rename Synopsys->Synopsis (I think that's the correct spelling if I'm not mistaken).


giograno:
yep, absolutely!

PR #204, opened 2022-07-13
fix lambda example filepath

bentsku:
Resolves https://github.com/localstack/localstack/issues/6448


bentsku:
The issue does not comes from this, we can close it

Issue #211, opened 2022-07-23
clarify that host mode is an unsupported developer utility


thrau:
we frequently receive questions (including but not limited to the one blow), about host mode not working as expected or not being documented correctly.

we shouldn't be advertising host mode as anything user facing in the first place. any reference to `--host` should move into the developer guide.

* #182

PR #213, opened 2022-07-25
move host mode docs to developer guide

HarshCasper:
Fixes https://github.com/localstack/docs/issues/211


thrau (approved):
thanks! this will solve so many headaches :relieved:

PR #218, opened 2022-08-01
switch from Lunr to Google Custom Search

### LINKED ISSUE #44: integrate online search
* algolia: https://www.algolia.com/
* in action: https://www.apollographql.com/docs/

alexrashed:
This PR enables the Google Custom Search feature for our docs.
We can verify in the preview if the search experience is better or worse than now :)

Fixes #44

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -123,13 +123,13 @@ github_project_repo = "https://github.com/localstack/localstack"
 github_branch= "main"
 
 # Google Custom Search Engine ID. Remove or comment out to disable search.
-# gcs_engine_id = "d72aa9b2712488cc3"
+gcs_engine_id = "35c306eb92f1549ca"
 
 # Enable Algolia DocSearch
 algolia_docsearch = false
 
 # Enable Lunr.js offline search
-offlineSearch = true
+offlineSearch = false
 
 # Enable syntax highlighting and copy buttons on code blocks with Prism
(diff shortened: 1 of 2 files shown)


HarshCasper (approved):

thrau (commented):
what bothers me a bit is that apparently also the sidebar nav is indexed, leading to search result showing pages where the AWS services nav is unfolded (see for example where SQS appears in PHP Docs and Boto Docs):

is this just a limitation or something we could fix?

image

alexrashed:
Interesting that Docsy doesn't handle that automatically: https://github.com/google/docsy/issues/261
The results for the GCS are the pretty much the same as for Google (at least for me): https://www.google.com/search?q=site%3Adocs.localstack.cloud+SQS
So it might make sense to fix that either way (even if we decide not to use the GCS).

It seems like the menu might need an attribute on the tag enclosing the menu to exclude the content from influencing the ranking.
Maybe something like `nosnippet`? https://developers.google.com/search/docs/advanced/robots/robots_meta_tag#nosnippet
Unfortunately, I think this will be some trial-and-error, because we need to change the tags on the production instance (or we could mess up our normal google search results)?

To be honest, for me the ads are pretty much a deal-breaker anyway (uBlock is quite good in blocking those 😄):
image

alexrashed:
Closed in favor of #221.

PR #220, opened 2022-08-04
switch from Lunr to Algolia DocSearch

alexrashed:
`docs.localstack.cloud` was accepted for Algolia's DocSearch program!

This PR is in direct competition to #218.
It integrates Algolia search instead of the current "offline" Lunr search.
The search does not have any Ads, has better results, and imho looks way better than the Google Custom Search:
image

You can play around with it in the preview deployment :)
Let me know what you think!


thrau (approved):
super cool! want!

HarshCasper (approved):

PR #221, opened 2022-08-04
switch to Hugo modules, upgrade docsy

alexrashed:
The integration of the Algolia search (#220) depends on a bugfix in Docsy from April.
Since we created the docs, both Hugo and Docsy added lots of changes.
This PR upgrades the repo to use the latest features of Hugo (namely Hugo modules) and the latest version of Docsy.

This has the following benefits:
- It introduces a clear versioning for Docsy (instead of the submodule commit hash reference).
- It re-ensures parity with the default project of Docsy in order to simplify upgrades in the future.


thrau (approved):
that looks great! thanks for the cleanup

PR #227, opened 2022-08-10
Minor: update IAM sample to allow request from root user

whummer:
Minor: update IAM sample to allow request from root user. Discovered this during testing / preparing an IAM sample - the command `awslocal s3 mb s3://test` actually succeeds with the new version of IAM, as by default we're running requests as the IAM root user. 👍


dfangl (approved):
Ah, what an oversight, thanks for fixing! Just sneaked in a minor typo change I found while checking your changes!

PR #231, opened 2022-08-17
auto-update parity coverage docs

alexrashed:
This automation has been created in collaboration with @steffyP.
The PR adds a GitHub workflow which is automatically executed every Monday morning.
It performs the following stages:
- Collect our LocalStack AWS parity coverage metrics from different test pipelines.
- Generate the coverage documentation (i.e. this page: https://docs.localstack.cloud/localstack/coverage/)
- Check if there are any changes to the current version.
- If so, it files a new PR or updates an existing PR to update the docs. @steffyP is configured as the default reviewer for the created PRs.

In later iterations, we should refactor the scripts such that they don't depend on LocalStack code anymore (since currently it's necessary to setup a localstack env in order to execute the scripts).


steffyP (approved):
LGTM 👍 🚀 
Thanks for adding the auto-update 🤖 😄

PR #259, opened 2022-09-12
DOC: Add usage tracking doc

HarshCasper:


content/en/localstack/usage-tracking.md
```diff
@@ -0,0 +1,117 @@
+---
+title: LocalStack usage tracking
+weight: 75
+description: >
+  Understand what data LocalStack collects and how you can opt out of usage tracking.
+---
+
+## Overview
+
```
whummer, line 9:
```suggestion
LocalStack is tracking a small amount of anonymized usage events, in order to optimize the user experience of our services. Please note that we always follow the _principle of least data_ - collecting only the essential information to provide the service, and only for the least amount of time possible or feasible. In general, only anonymized information is used, and we **never** collect any personal data or specific data about your application, unless you explicitly instruct the system to do so (in order for you to get additional application insights).

```

content/en/localstack/usage-tracking.md
```diff
@@ -0,0 +1,117 @@
...
+## LocalStack usage statistics
+
+For pro users, most of the information is collected to populate the "stack insights" dashboard. Collecting basic anonymized usage of AWS services helps us better direct engineering efforts to services that are used the most or cause the most issues.
+
+### Session information
+
+The current usage event collection on the client side includes:
+
+- A randomly generated ID pertaining to the session.
```
whummer, line 20: nit: We could unify the bullet points below, and either add a dot `.` at the end or remove it from all items.

content/en/localstack/usage-tracking.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+### AWS API call metadata
+
+The AWS API call metadata includes:
+
+- The service being called (like `s3` or `lambda`).
+- The operation being called (like `PutObject`, `CreateQueue`, `DeleteQueue`).
+- The HTTP status code of the response.
+- If it is a 400 error, we collect the error type and message. If it is a 500 error (internal LocalStack error), and `DEBUG=1` is enabled, we may also collect the stack trace to help us identify localstack bugs.
```
whummer, line 51:
```suggestion
- If it is a 400 error, we collect the error type and message. If it is a 500 error (internal LocalStack error), and `DEBUG=1` is enabled, we may also collect the stack trace to help us identify LocalStack bugs.
```

whummer (commented):
Thanks @HarshCasper for adding this 👍  - added a few minor comments/suggestions, would be great to add those before we merge..

Issue #240, opened 2022-08-23
Add docs for lambda event filtering


steffyP:
Create docs and some examples for new lambda event filtering feature: https://github.com/localstack/localstack/pull/6212

PR #262, opened 2022-09-12
DOCS: Add docs for lambda event filtering

HarshCasper:
Fixes https://github.com/localstack/docs/issues/240

### DIFF
--- a/content/en/aws/lambda/index.md
+++ b/content/en/aws/lambda/index.md
@@ -109,3 +109,37 @@ $ aws lambda add-layer-version-permission
 {{< / command >}}
 
 Next time you reference this layer (with the real AWS Lambda layer ARN) in one of your local Lambda functions, it will get automatically pulled down and integrated into your local dev environment.
+
+## Lambda Event Filtering
+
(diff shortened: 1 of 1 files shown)


content/en/aws/lambda/index.md
```diff
@@ -109,3 +109,37 @@ $ aws lambda add-layer-version-permission
...
+        {
+            ...
+            "dynamodb": { }
+            ...
+        }
+    ]
+}
+```
+
+If your DynamoDB event is being triggered, you can now apply additional logic via event filtering with patterns. The event filtering will use criteria checks mentioned above. However only five event filtering patterns is limited to a single Lambda function and each of the 5 patterns is validated against an OR condition. For more information, refer to the [official AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/invocation-eventfiltering.html).
```
steffyP, line 145: Actually LocalStack has currently no strict enforcement of 5 patterns implemented. However, as AWS enforces it, IMO it's ok to leave it here.

steffyP (approved):
LGTM 👍 thanks for adding the docs @HarshCasper 😄 

Only added a little remark, regarding the limit of filtering patterns that LocalStack currently doesn't enforce.

PR #263, opened 2022-09-12
add info about failing github runners

dominikschubert:
Adds info about the conflict between the two python libraries `pyopenssl` (when running an older version) and `cryptography`.

See https://github.com/localstack/localstack/pull/6831#issuecomment-1241974114


whummer (approved):
Thanks for adding this @dominikschubert ! 👍

PR #264, opened 2022-09-13
DOCS: Add docs to customize default IAM/STS user via config options

HarshCasper:
Adds docs as referenced by @whummer over the LocalStack Community repository: https://github.com/localstack/localstack/pull/5206


dfangl:
This will not work anymore, this functionality has been removed (as not in any way complient with AWS and interfering with IAM enforcing), which is why the documentation was removed

HarshCasper:
> This will not work anymore, this functionality has been removed in [localstack/localstack#6324](https://github.com/localstack/localstack/pull/6324) (as not in any way complient with AWS and interfering with IAM enforcing)

Thanks for sharing @dfangl. I missed it earlier 😅

PR #269, opened 2022-09-21
Add docs for EKS LB ports config and host-based routing

whummer:
* add docs for EKS load balancer ports config (via `_lb_ports_` cluster tags)
* add section and sample for host-based routing
* rename `__k3d_volume_mount__` to `_volume_mount_` and add deprecation notice


dfangl (approved):
LGTM!

PR #277, opened 2022-10-03
Remove multi-accounts feature flag

viren-nadkarni:
Multi-account will be baked-into the core system from v1.2 and cannot be disabled. This PR removes the defunct flag from the docs.


thrau (approved):
:+1: 

just noticing the challenge to keep the docs backwards compatible without versioning as we do localstack. i guess it's OK to keep the docs matched to only the latest version

HarshCasper (approved):

PR #278, opened 2022-10-03
docs: remove localstack terraform tests docs

HarshCasper:


thrau (changes requested):
i don't think we need to remove the entire section. just updating the description that the tests are run as part of our CI workflow, and that there is a github action that runs them.

PR #283, opened 2022-10-06
docs: add sns docs

HarshCasper:


bentsku (approved):
LGTM! Maybe we could add some informations about subscriptions with SQS queues, but then we'd start to intertwine different services in the docs, I'm not sure if that's a good thing.

bentsku:
Also, I'll need to add an entry about https://github.com/localstack/localstack/pull/6661 and the internal routes, where would be the right place for that in the docs?

thrau:
definitely this page @bentsku, maybe we can do a follow-up PR.

PR #289, opened 2022-10-07
Add SageMaker inference docs

silv-io:


thrau (approved):
thanks for adding the docs! i fixed a couple of the repo links. seems the training job sample is in the sample archive.

PR #291, opened 2022-10-08
docs: add docs for kms

HarshCasper:


taras-kobernyk-localstack:
The main question here is: what is the goal of this doc? There is a mention of "you'll learn how to make a basic usage of SNS", but no SNS is in the subsequent text. Maybe it is a leftover from an SNS documentation that was used as a template, but since KMS is used in other services, it might look like a tutorial on integrating SNS with KMS, so not sure.

Also the doc lists a lot of operations for KMS. If that is the goal - to list possibilities - maybe it is better to just point customers to AWS docs, as they are more complete and we do not have to worry about maintaining them.

The mention of our web interface is not justified, unfortunately. At least for KMS, that Resource page is not great. An example - it has a checkbox for multi-region keys. Had the mention before we even implemented multi-region keys in LocalStack. So the page requires an update before we can recommend it to people.

There are also some small things
- It not exactly clear from the tutorial, where the key ID comes from, it is just being used everywhere. So might be confusing to new people.
- We do not actually key key materiel rotation in LocalStack, even if we allow people to execute operations to enable/disable those rotations. So that is not the most useful functionality to teach to people - due to its absence.
- The tem " Customer Master Keys" is outdated according to AWS [docs](https://docs.aws.amazon.com/kms/latest/cryptographic-details/kms-keys.html), so its mention is confusing as well.

So what I propose is:
- Avoid mentioning the web app and Resources on pages for individual services. Simply because that app is not specific to certain services, rather a general tool, so might be better to have a page for it somewhere at the beginning of the documentation.
- Link the official AWS KMS docs from the page for the service. Less trouble for us with basics.
- Write instead a small tutorial that focuses on something, that might be considered to be a core of a service. In KMS it might be encoding/decoding of data using a key.
- Make our documentation as small as possible - to avoid confusing and intimidating readers. Might be better to not list all the possibilities than to scaring people away from using even simplest things.

If you are fine with that - I will write that tutorial beginning of next week.

And in generally - it might be better to delegate such things to service owners - simply because otherwise you are going to be too busy writing everything yourself. Also for KMS I know what works and what is not there. Wouldn't be fair to force you to figure that out for KMS or other services.

HarshCasper:
> There is a mention of "you'll learn how to make a basic usage of SNS", but no SNS is in the subsequent text. Maybe it is a leftover from an SNS documentation that was used as a template, but since KMS is used in other services, it might look like a tutorial on integrating SNS with KMS, so not sure.

Yes, I used a portion of the SNS doc and I missed out on removing those parts. They should be fixed now 👍 

> Also the doc lists a lot of operations for KMS. If that is the goal - to list possibilities - maybe it is better to just point customers to AWS docs, as they are more complete and we do not have to worry about maintaining them.

Unfortunately, most users don't refer to AWS docs. They try to look out for some basic commands and samples to try out first before moving to more complex use-cases.

> It not exactly clear from the tutorial, where the key ID comes from, it is just being used everywhere. So might be confusing to new people.

My bad! I had a note to remove them out before sending the PR. Will fix them now.

> We do not actually key key materiel rotation in LocalStack, even if we allow people to execute operations to enable/disable those rotations. So that is not the most useful functionality to teach to people - due to its absence.

Will remove those parts.

> Link the official AWS KMS docs from the page for the service. Less trouble for us with basics.

Will do.

> Write instead a small tutorial that focuses on something, that might be considered to be a core of a service. In KMS it might be encoding/decoding of data using a key.

That should be the purpose of a separate `Tutorials` section, and not service page which should just serve as a template for folks to start in a simplistic manner.

> The mention of our web interface is not justified, unfortunately. At least for KMS, that Resource page is not great.

I guess that should come up with the Web/Platform team. /cc: @whummer 

> Make our documentation as small as possible - to avoid confusing and intimidating readers. Might be better to not list all the possibilities than to scaring people away from using even simplest things.

I still don't see how the existing documentation is scaring people. Its purposefully geared towards first-time users to acquaint them with some basic commands before proceeding ahead with their more-complex use-cases (which a tutorial that you have proposed should cover).

> If you are fine with that - I will write that tutorial beginning of next week.

It would be better if that goes in the `beta` branch of the docs to accumulate service-focussed tutorials in one single place.

HarshCasper:
Superseded by: https://github.com/localstack/docs/pull/295

PR #294, opened 2022-10-11
Docs: Fix stream name in firehose example

tonisanta:
Hello, seems that in this snippet the stream name doesn't match with the rest


tonisanta:
Hi @dfangl , many thanks for your contribution, maybe you are interesed in reviewing this PR

dfangl (approved):
Thanks for the fix!

Issue #299, opened 2022-10-12
CloudFront example distribution ID missmatch


cmoralesmx:
The documentation for [CloudFront](https://docs.localstack.cloud/aws/cloudfront/) presents an example for creating a distribution to expose a file stored using an s3 bucket.
However, the distribution ID used in the description `abc123` and `https://abc123.cloudfront.net/hello.txt` does not match the distribution name used in the example `--origin-domain-name bucket1.s3.amazonaws.com`.
In the [AWS docs](https://docs.aws.amazon.com/cli/latest/reference/cloudfront/create-distribution.html#examples), both the description and the code use the same distribution ID

PR #300, opened 2022-10-12
Match distribution IDs on CF example

cmoralesmx:
Having mismatched distribution IDs can be confusing to some.
The AWS CLI docs do have matching distribution IDs on the description and the code of their example.
Issue #299


thrau (approved):
thanks @cmoralesmx !

cmoralesmx:
Thanks for accepting my contribution, @thrau. I just recently discovered this exciting project, but I'm looking forward to get more engaged.
By the way, could this PR be accepted for hacktoberfest, please?

PR #319, opened 2022-10-25
removing: Pulumi mobileanalytics endpoint from docs

aadi58002:
Updating the docs to remove the `mobileanalytics` endpoint for pulumi which no longer exists


HarshCasper (approved):

cphoover:
Hmmm after removing mobileanalytics `pulumilocal up` is now just running forever...

PR #320, opened 2022-10-26
feat: added og, twitter and other meta tags

ayushpaharia:
### Working on adding open graph for social share previews
- Added `meta` partial to add several meta tags [ titles, descriptions, links, images, languages ]

Issue [ #87 ]

@HarshCasper

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -82,15 +82,36 @@ weight = 1
 [outputs]
 section = ["HTML", "RSS"]
 
+enableEmoji = true
+enableGitInfo = true
+enableRobotsTXT = true
+canonifyURLs = true
+
 [params]
 copyright = "LocalStack"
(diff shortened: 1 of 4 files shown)


HarshCasper:
Hi @ayushpaharia — I would recommend commenting on the issue before putting up the PR. Anyways, I believe @Bisht13 won't have a problem with this.

thrau (changes requested):
thanks for the contribution!

it seems the meta tags aren't added to the pages correctly (you only included the meta tags in the redirect.html - which i would not touch at all).
image

there are also existing meta tags on the page that we should consider (some twitter and og tags are already set by the template).
i'm also a bit skeptical of images.weserv.nl, can we just reference images from our domain?

thanks!

PR #325, opened 2022-10-28
docs: Add nav buttons in footer. Make 'privacy policy' link more readable

### LINKED ISSUE #69: Docs readability enhancement: Adding Previous and Next pages Footer
Adding previous and next pages as footers for docs reading would enhance greatly the documentation readability.

snhardin:
Fixes #69 

* Adds navigation buttons in the footer to move to the previous and next pages in the section.
* Changes "Privacy Policy" link to be visible.

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -169,7 +169,6 @@ footer.footer-localstack {
 
     border-top: 1px solid #ced4da;
 
-    padding-top: 3rem;
     padding-bottom: 2rem;
 
     color: $gray-500;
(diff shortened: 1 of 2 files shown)


snhardin:
**With both previous and next links visible:**

footer with 2 links visible

**With no links visible:**

footer with no links visible

thrau (changes requested):
hi! thanks for the contribution. i think this is a nice idea, but it would be great if we could move the links into the page container rather than the footer. i think the links are not very visible in the footer, and i'm not sure they really belong there. maybe above the "last modified" at the bottom of the page container?

PR #337, opened 2022-11-07
LocalStack Beta Docs

HarshCasper:


steffyP (approved):
Great work!! 👏 🚀 
The new docs has a lot of improvements, and there is so much additional content! 😄 
I really like the new _Tutorials_ 🙂 

As all the PRs were already approved separately I only sanity checked the deployed docs once more - so far everything looks fine 👍

alexrashed (approved):
Awesome!
This changes the docs quite a lot, but it's now more structured and extensible. I verified that all the previous links are still working correctly (so the migration will be seamless, all links out there should still work).
This was quite an effort, especially because it took a bit (and long-living cross-cutting branches are always insanely hard to keep up to date). 🚀

PR #360, opened 2022-11-23
Rds docs update

steffyP:
add one sentence about RDS minor versions, and a note on the `engine_version` that is returned for db-cluster and db-instances


whummer (approved):
LGTM, great point about the Terraform compatibility 👍

PR #369, opened 2022-11-28
improve look of tutorial pages

lukqw:
image

Highlights:
- enables reading time for our pages, so all our pages now have an approx. indicator of how long it will take to read the article.
- this can be disabled (and has been disabled on quite some pages) with `hide_readingtime: true`
- introduce new layout for tutorials, which features last updated date, reading time, and a SM share button.
  - to use supply `type: tutorials` in content frontmatter


HarshCasper (changes requested):
Thanks a lot for working on this 🙏 

Some general-level stuff that I noticed:

- The share button across social media doesn't work as expected. If I try sharing it across Twitter or LinkedIn, it shares the URL path without the base URL (or is this expected to work after merge?). 
- The tutorials page lacks the brief description that we share as `linkTitle`. Can we add it so that the users get an understanding of what they can expect in the tutorial?

lukqw:
@HarshCasper thanks for trying out the changes. Pushed another fix that hopefully enables correct SM sharing.
For the `linkTitle`: This prop seems to be identical with the `title` over all tutorials - I assume you are referring to the `description`? 
Then it can look like this:
image

HarshCasper:
> @HarshCasper thanks for trying out the changes. Pushed another fix that hopefully enables correct SM sharing. For the `linkTitle`: This prop seems to be identical with the `title` over all tutorials - I assume you are referring to the `description`? Then it could look like this: image

It would look perfect then!

HarshCasper:
I tested the link sharing on LinkedIn and Twitter — The former is buggy while the latter works. I don't have a Facebook account to test suitably 😞

HarshCasper (approved):

PR #416, opened 2022-12-23
Cockpit 0.2.0

HarshCasper:


content/en/user-guide/tools/cockpit/index.md
```diff
@@ -26,63 +23,46 @@ On [Windows](https://www.windowscentral.com/how-disable-smartscreen-trusted-app-
...
 
 ## Known issues
 
-- MacOS ≤ v10 not supported yet.
-- There may be glibc issues on older Linux versions.
+- Mac OS X 10.0 and lower are not supported yet.
+- There may be `glibc` issues on older Linux versions.
 
-## Report issues
+While installing Cockpit on Mac M1, you may encounter an `xattr` error. To resolve this issue, you can visit the [Apple Developer Forum thread](https://developer.apple.com/forums/thread/692774).
```
thrau, line 61: technically it's not an "xattr error", mac os may prevent the software from running because of issues with code signing. to fix it, you can run `xattr -rc` in the terminal on the binary file and then run the application again.

thrau (approved):
LGTM! thank you! just a minor clarification with the mac software security would be useful. comment inline

PR #425, opened 2023-01-02
document sqs developer endpoints

HarshCasper:


thrau:
re-organized the sections to have the correct levels, and added example output

PR #448, opened 2023-01-25
Update IAM docs to use separate terminal sessions for admin commands

whummer:
Small update IAM in docs to highlight switching between different terminal sessions, as the administrative IAM commands need to be run with the default root user.

As an alternative, we could also keep it in a single code listing and add switches to temporarily set/unset the access credential environment variables. Feels like it is a bit more clear to have it separated into separate listings with two terminals (as in this PR). Thoughts @dfangl , which one makes more sense?

### DIFF
--- a/content/en/user-guide/aws/iam/index.md
+++ b/content/en/user-guide/aws/iam/index.md
@@ -8,54 +8,15 @@ aliases:
   - /aws/iam/
 ---
 
-By default, LocalStack uses not enforce security policies for client requests. In LocalStack Pro, the IAM security enforcement feature can be used to test your security policies and create a more realistic environment that more closely resembles real AWS.
-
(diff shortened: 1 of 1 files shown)


dfangl (commented):
I like it, but it refers to the terminals as "terminal 1" and "new terminal". The first "terminal 1" seems a bit sudden for me. As a user, I would think "what terminal 1, is there a second one?". Maybe we can add a little explanation beforehand, that two terminals are needed, and refer to them as "terminal 1" and "terminal 2" ?

whummer:
Great point @dfangl , totally agree that this was a bit confusing. Added some overall restructuring to the docs, and added clearer distinction between the two terminal sessions - please check again! 👍

dfangl (approved):
Love it, thanks for the restructoring, much easier to understand now!

PR #450, opened 2023-01-26
Add warning note about state incompatibility

viren-nadkarni:
image

---

image

### DIFF
--- a/content/en/references/persistence-mechanism.md
+++ b/content/en/references/persistence-mechanism.md
@@ -7,7 +7,10 @@ aliases:
   - /localstack/persistence-mechanism/
 ---
 
-## Overview
+{{< alert title="Warning" color="warning">}}
+Persisted states are only compatible with the specific version of LocalStack that was used to create them.
+The LocalStack team are actively working on addressing this limitation.
+{{< /alert >}}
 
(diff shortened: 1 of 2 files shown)


giograno (approved):
Thanks for jumping on this @viren-nadkarni 🙏 
Although we are oversimplifying here (because a pod can be compatible across versions if the stores and the moto backend do not change) probably this is a good way to cover ourselves here. What do you think @whummer?

whummer (approved):
Thanks for adding this @viren-nadkarni - looks good to me in general, and I think it's great that we're proactively informing our users about the pods/persistence limitations. 👍

I feel that we could reduce the severity of this message to a "note" (instead of warning), and we could also use more hedging language here - especially since we haven't established the clear semantics for compatibility across major/minor/patch versions yet.

Maybe we could use something along the lines of: _"Please note that there may be incompatibilities across different versions. Persisted states are currently only compatible with the version of LocalStack that was used to create them. We are actively working on addressing this limitation and will improve compatibility across versions in upcoming releases."_.

But no strong feeling - I'd also be fine with merging as-is.. 👍

viren-nadkarni:
@whummer Thanks, reduced the severity of alert box and rephrased the messages :+1:

PR #458, opened 2023-02-01
add moto coverage + refactoring of coverage docs generation

steffyP:
This PR:
* adds retrieval of the moto-test artifacts and includes the metrics collected in the coverage-overview.
* moves the coverage-creation script into docs repo, so we can remove the localstack dependency
  * we do rely on the content of `implementation_coverage_full.csv` (downloaded from circleci) to build the general model of available services/operations and then map the test results accordingly
  * renamed the raw-metrics data to include the information about the test-source
  * for now: add a different symbol to the tests that are only moto-tested
* adds a script to retrieve the correct github artifacts
  * Up until now we used the `gh run download --repo localstack/localstack-ext -n parity-metric-ext-raw` command, which would presumably download the latest artifact (from any branch). 
  * With the script, we try to download the latest artifact from the branch we are interested in (usually master), and if there is no artifact (because build steps have been skipped), we search for the related workflow that caused the skipping and try to download the artifacts from there.

**open TODOs:**
- [x] change branch-name and worflow name for the moto artifact after we merged https://github.com/localstack/localstack-moto-test-coverage/pull/1
   * latest run with resources from feature branch can be found here https://github.com/localstack/docs/pull/479
- [x] run the action to build docs again, to verify that the outcome is correct

**Discussion**
Do we want to use a different symbol to indicate a test have been covered by moto, vs. covered by our internal integration tests? /cc @HarshCasper @whummer @dominikschubert 

In this PR I proposed:
* ✨ to indicate covered by our integration test suite 
* 💫 to indicate covered by moto test suite


.github/workflows/docs-parity-updates.yml
```diff
@@ -14,19 +14,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout docs
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
alexrashed, line 17: Since you already upgraded the checkout action version, you could get rid of the last two remaining deprecation warnings with:
- Line 36: `actions/setup-python@v4` (no changes other than the version bump necessary)
- Line 86: Using `GITHUB_OUTPUT` instead of `set-output`:
  - `echo "diff-count=$(cat resources/diff-check.log | wc -l)" >> $GITHUB_OUTPUT` should do the trick.
  - Here's an example: https://github.com/localstack/localstack-terraform-test/blob/main/.github/workflows/main.yml#L26
  - Here's a very short blog post on the deprecation (with examples): https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/docs-parity-updates.yml
```diff
@@ -14,19 +14,12 @@ jobs:
...
-      - name: Checkout localstack (community)
-        uses: actions/checkout@v2
-        with:
-          repository: localstack/localstack
-          fetch-depth: 0
-          path: localstack
-
       - name: Set up system wide dependencies
         run: |
           sudo apt-get install libsasl2-dev jq wget
```
alexrashed, line 25: I think this step might not be necessary anymore?

dominikschubert (approved):
LGTM as well :)


Just an idea and not blocking anything here:

Regarding icons, another idea would be sort of "testing level"

1. externally tested (moto / terraform test suite / ?)
2. internally tested
3. internally tested, validated against AWS & involves snapshots

Can be encoded with icons in the future, but would be kinda nice for us internally as well to try to get to 100% level 3 coverage :grin:

thrau:
since this has been open for two week, i merged it now. if you have any comments @HarshCasper , please feel free to put an issue in the backlog so we can address it in a second iteration.

PR #492, opened 2023-02-21
Bump current activemq version

simonrw:
We updated the version in LocalStack to 5.16.6[1] but we did not change
the example.

[1]: https://github.com/localstack/localstack-ext/pull/1382


ackdav (approved):
Great catch - thank you 🚀

PR #493, opened 2023-02-21
Use the latest ubuntu runner

simonrw:
The 18.04 runner is being deprecated, and they have started adding [brownouts](https://github.com/actions/runner-images/issues/6002) disabling builds.


thrau (approved):
:rocket:

PR #500, opened 2023-02-27
Clarify port publishing needs for debugging purposes

dfangl:
## Motivation
Publishing the debugging port from a lambda container to the host is only necessary for VS Code debugging, not for Pycharm Professional. This was not properly represented and lead to confusion for the users.
Also, the limitations only applied to the VS Code version, as they are limitations of the port publishing.

## Changes
* Move some paragraphs / paragraph heading levels to make the needed configurations clearer.
* Changed `__local__` bucket key to `hot-reload`.


dominikschubert (approved):
LGTM :+1:

HarshCasper (approved):

PR #505, opened 2023-03-03
Coverage Docs v2

steffyP:
# Reworked Coverage Docs

## Restructuring of docs generation
* Templates hold the style and html structure
* Data templates (json-files) contain the content, that will be used and rendered when using the templates
   * script `create_data_coverage.py` creates the data-templates using the information from
       * implemented coverage (community/pro)
       * raw test metrics (community/pro; and other test sources) 
   * The MD are generated with basic information (MD template is in the script)
* another json-data file `service_display_name.json` was added with some additional service information
   * long-name/display name, short-name; link to api
   * used for the MD templates to display the full name (and make the search better)
   * currently static (needs to be adapted for new services) but nothing breaks if service is missing
* Section Overview
   * grid view with simple search  

## Update workflow:
  * latest run: https://github.com/localstack/docs/pull/515
  * will update the data-templates (json files), and add any MD that is still missing (if new services are added)

### TODOs

- [x] Coverage Docs moved to references, but added as second item, right below "Configuration" (one option that was discussed with @HarshCasper)
- [x] update links from "feature coverage"
- [ ] add links for each service to the coverage page -> will be part of https://github.com/localstack/docs/pull/415
- [x] remove old coverage docs

### Future Nice-to-haves
* pinned header for the table
* sortable table
* analytics events for "show tests"
* more information about tests
  * maybe include snapshot skipped paths
  * maybe origin information (internal/external operation call)
* logo/images for services + improved section overview


viren-nadkarni:
This is great!

Instead of having a dedicated section for coverage, any thoughts on moving the coverage pages as a children of the service doc page? Something like `/user-guide/aws/<service>/coverage`, and then have link in service doc to the coverage page.

thrau (approved):
this looks so cool already! much cleaner and easier to navigate than before! Absolutely love the "How to read the test details?" section, it really shows how you care about the UX :medal_sports:   i also really like the new table structure.

i think i only have one minor nitpick which is whether we could avoid showing the word "Coverage" in every title of the card? image design nits:
* It would be cool to make it visually a bit more appealing, like showing the [AWS architecture icon](https://aws.amazon.com/architecture/icons/) of the respective service in the overview.
* I wonder whether we can condense these columns by simply having a set if icons in the same cell, without having a column header. like the AWS icon for indicating that it's validated, appropriate icons for internal/external test suite, maybe the camera for the snapshot test. with a `title` text you can just hover over the icon to see what it means, or add it to the "how to read this table" with a list of the icons and what they mean. just a suggestion, but maybe out of scope of this PR :-)
  image


great work overall! :rocket:

steffyP:
> Instead of having a dedicated section for coverage, any thoughts on moving the coverage pages as a children of the service doc page? Something like `/user-guide/aws/<service>/coverage`, and then have link in service doc to the coverage page.

@viren-nadkarni: majority of feedback was a preference to a separate section where all the coverage-information is collected for the services. But Harsh and me will still discuss this.

> It would also be nice to define these terms:
> 
> ```
> Internal Test Suite | External Test Suite | AWS Validated | Snapshot Tested | Details
> ```

There is actually a section "[Terminology](https://localstack-docs-preview-pr-505.surge.sh/coverage/coverage_acm/#terminology)" under the table and it is linked from the meta-header `Testing*`. But I guess it's not that obvious 😅

steffyP:
I did some final changes including:
* removing the "Coverage" from the title -> it now uses the "real service name" as title
  * The title includes the abbreviated and the long-name of the service
  * The subtitle includes the api name 
  * Generally: the `md` files are created with the script `create_data_coverage.py` using the service-name information from `service_display_name.json` (this file needs to be maintained manually). If the service is not found in the json, we use the default, e.g. the api-name.

* changed the naming of `version` (indicating pro/community) in the table to `edition`
* moved the "LocalStack Coverage" into the `References` section, and added it as #2 right below `Configuration`.
  * As the coverage contain subpages, by default the title in the menu is bold (if not selected!). 
  * I am not sure if this is something we want to keep (I think it actually fits). But if not, I would need some helps to adapt the style here 🙏 
  * See screenshot here:

Screenshot 2023-03-20 at 11 10 32

* In theory the PR would be ready (after a formal review of the included script), not sure if we want to do this before the official release or not 🙂 

/cc @thrau @HarshCasper

Issue #511, opened 2023-03-11
[Improvement] SQS SNS subscription


Lewiscowles1986:
Share localstack SNS subscription via SQS, or point to AWS docs page.

I am willing to make this contribution (please signpost me)

thrau:
hey @Lewiscowles1986, do you mean you would like instructions in the docs on how to subscribe an SQS queue to an SNS topic?
i think that would be an excellent addition! if we can point to AWS docs that work against localstack, then that may be the easiest.

we'd appreciate the contribution

Lewiscowles1986:
Thanks @thrau. The AWS docs I am aware of are less clear about that; but I've submitted a draft, which I hope gives us something to iterate upon, or submit.

PR #512, opened 2023-03-11
docs: SNS + SQS integration notes

Lewiscowles1986:
Proposal to Close https://github.com/localstack/docs/issues/511


Lewiscowles1986:
Question. Is there a way to view the preview myself?

I tried to follow conventions in the existing document, but I don't actually know how to preview this.

Lewiscowles1986:
I have now checked locally. It looks to be the same style.

It does take from 4 minute estimated reading to 5 minute estimated reading.

This is only my second foray into Hugo, so if it needs to come out to move elsewhere I might need support with what that would look like.

My passionate plea is that quite a lot of third-parties cover SNS + SQS, but the pieces are not that clear. By signposting someone in your docs, you enable them to both complete the task, but also understand the high-level steps.

bentsku (commented):
Hi @Lewiscowles1986 and thanks for your contribution!

It's definitely a great idea to add more samples and ways to get started with SNS and SQS.
I'd like to also have the responses from the command, so it would be a bit easier to read and follow where the resources come from (the ARN, the queue URL...).

I'd like to have your opinion @HarshCasper on the format.

On the content level, we maybe should set a Policy on the queue so that the SNS topic is able to publish to it. It is not enforced by LocalStack at the moment, but it could be good to mention it and maybe show how it should be done so that the sample is also working against AWS (maybe with a comment that it's not necessary? What do you think @HarshCasper? Maybe mentioning it and linking to https://docs.aws.amazon.com/sns/latest/dg/subscribe-sqs-queue-to-sns-topic.html would be enough.)

thrau:
since there hasn't been any movement on this, i took the liberty of pushing some changes that also @bentsku suggested, resolved merge conflicts, and will merge this once green. thanks for the contribution Lewis @Lewiscowles1986, and sorry we've been sitting on this for so long, it somehow slipped through the cracks

Lewiscowles1986:
@thrau I'm also sorry, as I think I forgot to make some changes.

PR #520, opened 2023-03-17
Rename default container name from localstack_main to localstack-main

simonrw:
Docker can use a container name for DNS resolution, but hostnames cannot include underscores.

This change sets the default container name in the documentation to `localstack-main` to allow it to be used as a hostname.

### DIFF
--- a/.github/workflows/markdown.links.config.json
+++ b/.github/workflows/markdown.links.config.json
@@ -34,6 +34,9 @@
             "pattern": "http://localstack_main:4566"
         },
         {
+            "pattern": "http://localstack-main:4566"
+        },
+        {
             "pattern": "(.*)localhost.localstack.cloud(.*)"
         }
     ],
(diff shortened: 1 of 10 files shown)


joe4dev:
Would be great to update the [docker-compose.yml](https://github.com/localstack/localstack/blob/master/docker-compose.yml) and docker-compose-pro.yml as well in the localstack repo. They are linked from the docs and often used as a template. Therefore, they spread everywhere :)

simonrw:
> Would be great to update the [docker-compose.yml](https://github.com/localstack/localstack/blob/master/docker-compose.yml) and docker-compose-pro.yml as well in the localstack repo. They are linked from the docs and often used as a template. Therefore, they spread everywhere :)

Hi, yeah my plan was to do this across the board, tackling every repo, but I did not complete this work before V2. Since it's a significant change and the problem it's fixing is relatively minor, I decided to not continue with this work right now.

PR #522, opened 2023-03-18
update persistence docs

thrau:
This PR updates the [persistence documentation](https://localstack-docs-preview-pr-522.surge.sh/references/persistence-mechanism/) to the current version that uses our new persistence plugin.

There are still some TODOs left to document the current limitations of services


giograno (approved):
👍 we should keep adding limitations while we keep increasing the coverage across all the services.

PR #523, opened 2023-03-19
add tutorial on deploying container service on ecs and ecr

HarshCasper:


thrau:
it seems 4 reviewers may not be necessary, so i removed myself :-) it would probably also be better to involve service owners in tutorials that concern them - i know very little about ECS or ECR

content/en/tutorials/ecs-ecr-container-app/index.md
```diff
@@ -0,0 +1,553 @@
...
+
+```dockerfile
+FROM nginx
+
+ENV foo=bar
+```
+
+The `Dockerfile` uses the official `nginx` image from Docker Hub, enabling us to serve the default index page. Before building our Docker image, we will start LocalStack and create an ECR repository to push our Docker image. Start LocalStack with the `LOCALSTACK_API_KEY` environment variable to use your LocalStack Pro API key.
+
+```bash
```
dfangl, line 36: Please use the `{{ command }}` shortcode to prevent people selecting the `$` when copy-pasting. This is applicable on every ```bash on this page.

content/en/tutorials/ecs-ecr-container-app/index.md
```diff
@@ -0,0 +1,553 @@
...
+$ LOCALSTACK_API_KEY=<your-api-key> localstack start -d
+```
+
+Next, we will create an ECR repository to push our Docker image. We will use the `awslocal` CLI to create the repository.
+
+```bash
+awslocal ecr create-repository --repository-name <REPOSITORY_NAME>
```
dfangl, line 43: Here, the $ prefix is missing as well.

dfangl (changes requested):
Some concerns about validity of statements / not-used shortcuts.
I haven't actually tried the sample against LS, I assume someone already did :) 
BTW: In general, for markdown + VCS, it's often recommended putting each  sentence on a single line (as only double linebreaks are linebreaks in the rendered markdown anyway). That would have made reviewing / commenting on single lines a tad easier.

PR #526, opened 2023-03-22
Update configuration document with new networking description

simonrw:
The new variables require explanation on how they work, and what the default behaviour will be if they are absent.

See the [implementation PR in the community repo](https://github.com/localstack/localstack/pull/7893)


thrau:
hey @simonrw, what's happening with this PR?

PR #536, opened 2023-03-27
fix dead link to parity-testing details

steffyP:
The link to `parity-testing` was relative, and pointing to a non-existing page. 

Changed the link to use the path from root.


silv-io (approved):
LGTM!

PR #543, opened 2023-03-29
remove docs on DISABLE_TERM_HANDLER

alexrashed:
Removes a small entry in the configuration reference documenting the `DISABLE_TERM_HANDLER` environment variable. This variable will be removed with 2.0.
This flag has just been introduced to document the changed default behavior (but this "new" default behavior is actually just to correctly handle signals).


dfangl (approved):
:killitwithfire:

PR #559, opened 2023-04-05
change upvote feedback to not include actionable item

lukqw:


ackdav (approved):
Thx, I already find this more intuitive. it might also be good to link to https://localstack.cloud/community/get-involved/ here. since the user is willing to participate, he/she might also be willing to get involved more

lukqw:
@ackdav thanks for the feedback! I have adjusted the link accordingly.

PR #568, opened 2023-04-10
add docs for localstack logging

HarshCasper:


content/en/references/logging.md
```diff
@@ -0,0 +1,58 @@
...
+  Overview of LocalStack logging and error reporting
+---
+
+LocalStack supports logging output and error reporting through the `localstack` CLI or a Docker/Docker Compose based setup. LocalStack's logging setup allows you to:
+
+- Discover errors in your code during development & testing.
+- Get visibility into how and why your API calls are failing.
+- Figure out unexpected errors such as Lambda timeouts and more!
+
+With LocalStack logging, you can easily retrieve additional detail around errors using various configuration variables to specify the verbosity and the log level.
```
whummer, line 16:
```suggestion
With LocalStack logging, you can easily retrieve additional details around errors, using various configuration variables to specify the verbosity and the log level.
```

content/en/references/logging.md
```diff
@@ -0,0 +1,58 @@
...
+
+`DEBUG` can be either `0` or `1` (`0` is the default). With `DEBUG`, you can print more verbose logs, useful for troubleshooting issues. `LS_LOG` supports the following values:
+
+- `trace`
+- `trace-internal`
+- `debug`
+- `info`
+- `warn`
+- `error`
+- `warning`
```
whummer, line 30:
```suggestion
- `warn`/`warning`
- `error`
```

whummer (commented):
Thanks for adding this @HarshCasper , only tiny nit comments..

Would be good to get a quick look from @thrau @dfangl as well who have been working on the logging refactoring, to see if we're missing anything.. 👍

content/en/references/logging.md
```diff
@@ -0,0 +1,57 @@
...
+`DEBUG` can be either `0` or `1` (`0` is the default). With `DEBUG`, you can print more verbose logs, useful for troubleshooting issues. `LS_LOG` supports the following values:
+
+- `trace`
+- `trace-internal`
+- `debug`
+- `info`
+- `warn`/`warning`
+- `error`
+
+The `LS_LOG` overrides the `DEBUG` variable and will provide the same log format but append the request and response objects and the HTTP headers to the log line. With `DEBUG=1`, errors inside LocalStack are reported to the client in full, and these stack traces can help you better triage your issues.
```
dfangl, line 31: I feel this part here is unclear. The relationship between LS_LOG and DEBUG is not well defined, sadly. In general: LS_LOG affects the log handlers level directly. Also, if LS_LOG is trace or trace-internal, it will automatically set DEBUG=1 as config as well.

So, if DEBUG=0 and LS_LOG=debug, the log handlers will behave on debug level, but any code checking the DEBUG variable directly in the config, will still think DEBUG=0. An example for this is the stacktrace returned in the service response, this would not be included then.

This is confusing, so I can't really say how to best include it in the docs, but I would suggest, if you really want debug information about LS, set DEBUG=1 explicitly as well.

My second issue with this sentence is `The `LS_LOG` overrides the `DEBUG` variable and will provide the same log format but append the request and response objects and the HTTP headers to the log line.`. It will only append this when the LS_LOG levels are trace or trace-internal, as well as DEBUG is only overridden there. 

All in all, I feel a little bit more information is required here.

dfangl (changes requested):
There is some unclarity in the explanation for LS_LOG, other than that, everything seems good, thanks for this!

PR #585, opened 2023-04-18
add docs for cdk for terraform on localstack

HarshCasper:


whummer (approved):
Looks great, kudos for adding this @HarshCasper ! 🚀 

Would be great if we provide a `cdktflocal` command (had that in the back of my mind for some time already), to further simplify the configuration. Maybe a topic for an upcoming hackathon.. :)

PR #587, opened 2023-04-19
add a note on api gateway custom id creation

HarshCasper:


bentsku (approved):
LGTM! Thanks for adding the note on this based on the support case 👍

PR #589, opened 2023-04-19
add docker-in-docker note for gitlab ci

HarshCasper:


MarcelStranak:
I believe that we should continue with the previous example and add another one for Docker in Docker.
@macnev2013, @dominikschubert what do you think? I'm aware that you had more experience with GitLab's CI. 

Could we also mention that there could be differences between public runners and their private runners? Furthermore, different CPU architectures can have an impact.

content/en/user-guide/ci/gitlab-ci/index.md
```diff
@@ -25,30 +25,46 @@ We recommend taking the following steps:
...
-      alias: localstack
+    - name: docker:20.10.16-dind
+      alias: docker
+      command: ["--tls=false"]
+
   before_script:
-    - LOCALSTACK_URL=http://localstack:4566
-    - pip install awscli
+    - apk update
+    - apk add gcc musl-dev linux-headers py3-pip python3 python3-dev curl
```
dominikschubert, line 51: curl isn't being used so it can be dropped

MarcelStranak (commented):
lgtm

dlsniper:
Hi. Sorry to comment on a closed PR, but I have a question regarding the change.
Why is it better to run `dind` as a GitLab service and then use Localstack with `docker pull` rather than configuring Localstack as a service in GitLab?
I couldn't find any explanation about this in the documentation.
I ask because I did just the reverse of what the documentation says, and instead of using `dind`, I made Localstack a service.
Thank you.

PR #594, opened 2023-04-21
add docs on aws sdk for ruby

HarshCasper:


MarcelStranak:
Does what it says it will do. I would also like to add the result that you should receive if it was successfully completed.

Result:
```
$ ruby .\localstack.rb                                                                                                                            ─╯
Created bucket doc-example-bucket-b911f85f-4dd3-4668-a32e-3f69aa4e37dc.
Your bucket's region is: us-east-2
```

PR #608, opened 2023-04-29
add loan broker application

HarshCasper:


steffyP (changes requested):
Generally, LGTM 👍 😄  very happy to see the new sample in the developer hub. 

**Only thing that still needs to be addressed**: there is no svg for the service SFN? not sure if this should have another name, or the svg just need to be created + added. 
image

Also we should think how to promote the `tags` - it seems like only two samples (this included) currently list the included services as `tags`, while the others seem to focus on broader categories.

tinyg210:
i just added the icon for step function.
and i agree, the tags with individual services are a little too much. we already have the icons, which will have tooltips soon and the title also contains all the services used.

steffyP:
Thanks @tinyg210 👍 I just realized that the README states that an API key is required, but there is no tag for PRO in the docs. Not sure if it requires PRO, but either the README or the docs should be adapted accordingly 🙂 

Regarding the double-service tags: please decide how you want to consolidate this @tinyg210 @HarshCasper 🙏  
Doesn't have to be tackled with this PR. 
If you could just clarify the requirement for PRO in this sample it's ready to be merged IMO 🚀

tinyg210:
i tested the app using only the community edition and everything works as expected (including cloud pods). i adjusted the REDAME file accordingly.

tinyg210 (approved):

PR #614, opened 2023-05-05
Support for OpenSearch 2.5

viren-nadkarni:
Added in https://github.com/localstack/localstack/pull/8167


alexrashed (approved):
Thanks for adding the new version to the docs! 🚀

Issue #617, opened 2023-05-05
Make Lambda layer docs a little simpler


lizconlan:
Having managed to get myself a bit lost trying to use a freshly created Lambda layer to a create-function command, it feels like it would be simpler to modify the jq query to fetch the `LayerVersionArn` (the `LayerArn` with the version number already appended) rather than manually adding it as part of the call to `lambda create-function`.

It's also a bit easier to follow what's happening across 2 different commands.

ackdav:
Hi @lizconlan - I'm closing this due to the linked PR already having been merged (appreciate your input! :rocket:). If you feel like there's still room for improvements, feel free to open a new one :smiley: thx!

PR #618, opened 2023-05-05
Simplify lambda layer creation

lizconlan:
This  PR proposes using `LayerVersionArn` instead of manually appending `:1` to the ARN to simulate a version number (missing it off results in a regex match failure message, causing `lambda create-function` to fail; you can guess how I know this!)

Not directly relevant for the current examples, but using the versioned ARN string will also simplify updating a layer and running `update-function-configuration` as it should provide the reference for the freshly updated layer with no further effort for the user.

Related issue: https://github.com/localstack/docs/issues/617


dfangl (approved):

dfangl:
LGTM, thank you for your contribution!

PR #654, opened 2023-05-31
Fix Configuration `INIT_SCRIPTS_PATH` Description

aroxby-wayscript:
On the page, https://docs.localstack.cloud/references/configuration/ This section:
> _Deprecated_. Specify the path to the initializing files with extensions `.sh` that are found default in `/docker-entrypoint-initaws.d`.

Should read:
> _Deprecated_. Specify the path to the initializing files with extensions `.sh` that are found default in `/etc/localstack/init/ready.d`.

This was confusing for me as the [init-hooks](https://docs.localstack.cloud/references/init-hooks/) reference page is not as easily found.

Alternatively, the docs for `INIT_SCRIPTS_PATH` could link to the init-hooks docs.


thrau (changes requested):
hey Andy @aroxby-wayscript! Sorry this fell through the cracks. The configuration.md has changed in the meantime (we moved all deprecated variables into their own section), and I think we should remove this variable all together, or at least mark it as defunct. It's not just deprecated, its not being used any more at all since you cannot configure the path to the init hooks anymore. 

So maybe something like

`**Deprecated**. Before 1.0, this was used to configure the path to the initializing files with extensions `.sh` that were found in `/docker-entrypoint-initaws.d`. This has been replaced by the init-hook system` (and link to the init hooks)

aroxby-wayscript:
Thanks for the update @thrau!  I'd like to keep this variable in the list and include a link to the init hook system.  Having this link here would have saved me a little bit of a headache when I was trying to get setup.

thrau (approved):
Thanks a lot!

PR #663, opened 2023-06-19
fix script for docs update

steffyP:
Latest [failing workflow for updating the docs](https://github.com/localstack/docs/actions/runs/5308007605/jobs/9607061813), revealed that community results might have operations included, that pro doesn't have yet. This could be as we consider the latest successful master build, and the repositories are split.

Fixed the KeyError issue, by setting defaults.


steffyP:
[Tested action successfully](https://github.com/localstack/docs/actions/runs/5309114420/jobs/9609357941)

HarshCasper (approved):

PR #667, opened 2023-06-19
Fix Lambda debugging instructions for Python listen

joe4dev:
When running a Lambda in a Docker container, we need to listen to any host and not only to `127.0.0.1` by default (see [listen](https://github.com/microsoft/debugpy/blob/main/src/debugpy/public_api.py#L94) implementation).


dfangl (approved):
Good catch!

PR #677, opened 2023-06-27
Improved create-applications script for developer hub

webdev51:
In developer hub, we have applications page which are coming from `data/developerhub/applications.json` file but as these are just JSON objects and not actual pages, we were facing some issues like it wasn't shown in the search + it was not very SEO friendly.

To resolve it, we created phantom pages which just redirects to the application itself but have some meta data so the pages are searchable now and SEO friendly as well. 

To create these pages automatically from `applications.json`, a script was written that was called `create-applications.sh` but it had some dependencies like `jq` so its moved to node.js (with no dependencies) and now the script lies in `scripts/create-applications.js` with very simple usage i.e. `node create-applications.js` and in result it'll create all the phantom pages with right tags and metadata. 

So whenever we update `applications.json` file. we'll have to run `node create-applications.js` to create the phantom pages.

The same has been updated in the readme file as well


lukqw (approved):
nice job 🚀

PR #679, opened 2023-06-30
mark DEBUG=0 as default

lukqw:


thrau (approved):
much clearer now! :+1:

Issue #646, opened 2023-05-26
Add documentation about CFn Intrinsic functions


pinzon:
There is currently no documentation for the cfn intrinsic functions in the Localstack documentation. This makes it difficult for developers to learn how to use these functions, and it can lead to errors in their code.

I propose that we add documentation for the cfn intrinsic functions to the Localstack documentation. This documentation should include:

1. A list of all the cfn intrinsic functions that are supported on Localstack
2. A description of how each function works
3. Examples of how each function can be used

PR #681, opened 2023-07-03
add template features explanation

pinzon:
This PR adds a more detailed list of the template features that our current implementation of Cloudformation supports.

Addresses: #646


alexrashed (approved):
LGTM! 🚀

PR #326, opened 2022-10-28
docs: Add documentation with example for LocalStack DocumentDB

snhardin:
Fixes #111 

Adds a documentation section for LocalStack DocumentDB.

PR #685, opened 2023-07-06
add documentdb documentation started in a previous PR

tinyg210:
Enhanced docs started in https://github.com/localstack/docs/pull/326. Contains:
- create cluster
- create database
- describe cluster
- mongosh interacting with databases
- creating collections in databases


content/en/user-guide/aws/docdb/index.md
```diff
@@ -0,0 +1,230 @@
...
+      }
+```
+
+Some noticeable fields:
+
+- `--db-instance-identifier test-company`: Represents the unique identifier of the newly created
+  database
+- `--db-instance-class db.r5.large`: Is the instance type or class for the Amazon DocumentDB
+  instance. It determines the compute
+  and memory capacity allocated to the instance. `db.r5.large` refers to a specific instance type in
```
tinyg210, line 123: @HarshCasper help me out here, is this `db instance class` mocked or ignored? is it relevant to describe it in any way?

steffyP: the `db-instance-class` will be returned for the `describe-db-instance`, but it doesn't change the behavior. So I would argue it's both: mocked + ignored (because we don't change the actual deployment based on this setting). 

Personally, I think it's not relevant for the docs, as we currently do not have any parity checks for this parameter. On AWS certain configuration combinations are not allowed (especially when it comes to cluster + instance combinations).

steffyP (approved):
Thanks for adding docs for docdb, @tinyg210 🙏 
IMO we could merge this 👍 
However, as we do not really make use of the `db-instance-class` in LocalStack, I think it's not necessary to go in such detail on the parameter specifics.

tinyg210:
hey @steffyP, I fixed the issues of including the original author. Initially, I was concerned that the folder structure had changed too much and the cherry-pick would make it hard to fix that, but upon closer inspection, it wasn't difficult at all. I didn't realize I forgot to give credit.

PR #695, opened 2023-07-14
improving docs for persistence endpoints

giograno:
Add info about the snapshot endpoints that do a complete save/load operation.


thrau (approved):
thanks for adding!

PR #700, opened 2023-07-17
coverage table enhancement for lambda

steffyP:
Skip the "external-test-suite" column for lambda in the coverage-overview, as we do not have any external tests.

Initially the column for "external tests" was added to show an improvement of the coverage. Mainly for services where we rely on moto, and which are therefore covered by moto-tests. 
However, for lambda this does not make sense, as the external test suite is not working, but do we have all tests covered internally (which is the end goal anyhow).

As the table looked now lambda is missing tests (which it doesn't), we decided to take the approach of excluding the "external test suite" for now with @joe4dev. Other services could be added to the exclusion. 

/cc @HarshCasper @whummer

### DIFF
--- a/layouts/partials/coverage/coverage_table.html
+++ b/layouts/partials/coverage/coverage_table.html
@@ -1,3 +1,4 @@
+{{ $service := $.service }} 
 <div class="coverage-report">
     <table>
         <thead>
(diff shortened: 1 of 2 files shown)


steffyP:
https://localstack-docs-preview-pr-700.surge.sh/references/coverage/coverage_lambda/

joe4dev (approved):
Works for now making the table look better 👍 

We can extend/modularize this if it becomes relevant for other services as well with an exclude list.

PR #708, opened 2023-07-19
update expected health endpoint response

lpmi-13:
the existing example in this part of the docs shows the result of the default `opensearch_domain.com/` endpoint rather than the `opensearch_domain.com/health` endpoint, so I updated it to show the same response as exists earlier in the same page.

### DIFF
--- a/content/en/user-guide/aws/opensearch/index.md
+++ b/content/en/user-guide/aws/opensearch/index.md
@@ -449,21 +449,22 @@ $ awslocal opensearch describe-domain --domain-name my-domain
 {{< command >}}
 $ curl my-domain.us-east-1.opensearch.localhost.localstack.cloud:4566/_cluster/health | jq
 {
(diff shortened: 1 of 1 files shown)


lpmi-13:
if, on the other hand, the intent here was to show the response from the default domain path (instead of the `/health` endpoint), I can easily revert the response change and instead update the path that's being used in the request.

alexrashed (commented):
Hi @lpmi-13!
I am really sorry for the late answer. Thanks a lot for your contribution! Unfortunately, this documentation has been changed a bit in the meantime with https://github.com/localstack/docs/pull/761 (which fixed this section).
Please let us know if you can still find any issues on the OpenSearch (or any other) page in the docs.

lpmi-13:
No worries! Love this project, and keep up the great work!

giphy(24)

PR #719, opened 2023-07-24
Added copy button for code snippets

webdev51:
Added copy button for code snippets which will copy code snippets to clipboard

image


static/js/global-script.js
```diff
@@ -0,0 +1,18 @@
...
+      $(this).find('span').text('Copy');
+    }, 2000);
+  })
+}
+
+$('.highlight pre').each(function(_i,item){
+  $(item).addClass('code-container');
+  $(item).append('<button class="code-copy-button"><i class="fa fa-copy"></i><span>Copy</span></button>')
+})
+codeCopyBtnClickHandler();
```
lukqw, line 18: do we need the separate method definition, since we're defining an on click handler in there?
seems to me like we could just do 

```
$('.code-copy-button').on('click',function(e){
    e.preventDefault();
    const elem = $(this).closest('.code-container').find('code').clone();
    elem.find('.command-prefix').remove(); // removing prefix
    navigator.clipboard.writeText(elem.text());
    $(this).find('span').text('Copied!');
    setTimeout(() => {
      $(this).find('span').text('Copy');
    }, 2000);
  })
```

here

do we also wanna handle the `no-copy-part` here as well which we briefly talked about?

webdev51: The idea behind separating the definition and call part is for future. As the file name suggests, its a global javascript file (as there wasn't any previously) and in future, we might want to add some more global functions so it'd look like below:
```
function codeCopyBtnClickHandler(){...}
function a(){...}
function b(){...}
function c(){...}

codeCopyBtnClickHandler();
a();
b();
c();
```

About the `no-copy` div, if we pass a span in a pre block, instead of rendering it, its showing it as it is like its a part of the command itself. So it's something we've to find an alternate solution but I thought that its not that crucial that we hold this copy button. I can create a new issue in which I can mention the issue and share it in some sig channel most probably so everyone updates their doc pages?

webdev51: Good call, done!

lukqw (approved):
noice 🪂

PR #724, opened 2023-07-25
update rds using secrets

steffyP:
updated sample to use secrets in json-format


whummer (approved):
Thanks for updating this @steffyP ! 👍

PR #725, opened 2023-07-25
enhance docs for IAM

HarshCasper:


dfangl (approved):
I think this looks good! However, Osama is currently reworking the IAM resource browser, not sure if there will be visual changes.

However, we should really stick to the one-sentence-per-line rule

PR #730, opened 2023-07-27
add java sdk getting started examples and resources

tinyg210:


content/en/user-guide/integrations/sdks/java/index.md
```diff
@@ -0,0 +1,245 @@
...
+
+    // The key of the object in the bucket, usually the name of the file.
+final String key = "hello-v1.txt";
+
+    // S3 Client with configured credentials, endpoint directing to LocalStack and desired region.
+AmazonS3 s3Client = AmazonS3ClientBuilder.standard()
+.withCredentials(new AWSStaticCredentialsProvider(credentials))
+.withEndpointConfiguration(new EndpointConfiguration("s3.localhost.localstack.cloud:4566",
+Regions.US_EAST_1.getName()))
+.build();
```
thrau, line 53:
```suggestion
AmazonS3 s3Client = AmazonS3ClientBuilder.standard()
  .withCredentials(new AWSStaticCredentialsProvider(credentials))
  .withEndpointConfiguration(new EndpointConfiguration("s3.localhost.localstack.cloud:4566", Regions.US_EAST_1.getName()))
  .build();
```

indenting the function chaining makes the code a bit easier to read. maybe we could add this everywhere

thrau (commented):
:100: so glad we're finally adding this!

* maybe we could introduce H3 sections `### Interacting with S3``, `### Interacting with Dynamodb`. something like this, to visually separate the two examples? makes it easier to read
* does it make sense to maybe introduce before the example section a general section on `## Client configuration`, to give a general intro in how to configure Java SDK clients to connect to localstack? So explain the endpoint, test credentials, etc. On that note: are there any other interesting/relevant things to configure in the clients, or ways to configure them?

content/en/user-guide/integrations/sdks/java/index.md
tinyg210: I added the indentation and separation of client configuration and interaction with the services. I also made a clearer separation of sections based on services. As for the client configs, there are some options to add special handlers and some monitoring but that is really niche and should be covered in the official docs.
I think the configs in general are pretty self-explanatory 🤔

thrau (approved):
lgtm! :coffee:

PR #731, opened 2023-07-28
add section to sqs on accessing queues from lambdas

thrau:
This PR addresses #627. Generally this is something that users have been struggling with, and I think this initial iteration should clarify most of the questions people have.


simonrw (approved):
This is a big improvement, and is nice and clear on what to do. We still have the issue of connecting from both lambda _and_ docker container, but we are working on that.

LGTM :+1:

PR #742, opened 2023-08-02
Added LocalStack Academy

webdev51:


viren-nadkarni (approved):
Great stuff @Nancy-Chauhan!

Overall the content and recordings are excellent. I only have a few nitpicks/suggestions.

I strongly recommend using one sentence per line in all the markdown files.

Also, I felt the language was a bit too marketing heavy, with too many instances of words like 'revolutionary' or 'exciting' or 'game-changing'. We might want to tone it down it future iteration since this is intended to be educational in nature :)

PR #743, opened 2023-08-02
add NodeJS config to "Custom TLS certificates"

alexrashed:
This PR adds the NodeJS Custom CA cert config to the ["Custom TLS Certificates"](https://docs.localstack.cloud/references/custom-tls-certificates/) docs page.
This is necessary, since we use NodeJS for some services, for example to download the `kinesis-mock` package using `npm`.

### DIFF
--- a/content/en/references/custom-tls-certificates.md
+++ b/content/en/references/custom-tls-certificates.md
@@ -23,8 +23,9 @@ There are three options when running LocalStack:
 They all can be summarised as:
 
 1. get your proxy's custom certificate into the system certificate store, and
-2. configure [`requests`](https://pypi.python.org/pypi/requests) to use the custom certificate, and
-3. configure [`curl`](https://curl.se/) to use the custom certificate.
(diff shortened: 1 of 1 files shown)


ackdav (approved):
Looks good to me, the only thing that confused me slightly was the naming. We call it `_CA_BUNDLE` for curl and requests, but for node it's `EXTRA_CA_CERTS`? I probably lack the background knowledge why a `ca_bundle` isn't appropriate within node...

alexrashed:
These are environment variables defined by the different tools:
- [`requests`](https://requests.readthedocs.io/en/latest/user/advanced/?highlight=REQUESTS_CA_BUNDLE#ssl-cert-verification) uses `REQUESTS_CA_BUNDLE`.
- [`curl`](https://curl.se/docs/sslcerts.html) uses `CURL_CA_BUNDLE`
- [`node`](https://nodejs.org/api/cli.html#node_extra_ca_certsfile) uses `NODE_EXTRA_CA_CERTS`

PR #764, opened 2023-08-15
enhance docs for rds

HarshCasper:


content/en/user-guide/aws/rds/index.md
```diff
@@ -1,83 +1,44 @@
...
-Snapshots are currently not supported for MariaDB.
-
-### MySQL Engine
-
-By default, a MariaDB installation is used when requesting a MySQL engine type. 
-
-If you wish to use a real MySQL version, you can do so by setting the environment variable `RDS_MYSQL_DOCKER=1`. With this feature enabled, MySQL community server will be started in a new Docker container when requesting the MySQL engine. The `engine-version` will be used as the tag for the image, meaning you can freely select the desired MySQL version that is listed on the [official MySQL Docker Hub](https://hub.docker.com/_/mysql).
-
-In case you want to use a special image, you can also set the environment variable `MYSQL_IMAGE=<my-image:tag>`.
+RDS (Relational Database Service) is a managed database service provided by Amazon Web Services (AWS) that allows users to setup, operate, and scale relational databases in the cloud. RDS allows you to deploy and manage various relational database engines like MySQL, PostgreSQL, MariaDB, and Microsoft SQL Server. RDS handles routine database tasks such as provisioning, patching, backup, recovery, and scaling.
```
steffyP, line 10: not sure it's a good idea to outline this here, because LocalStack doesn't support this currently:
> backup, recovery, and scaling

content/en/user-guide/aws/rds/index.md
```diff
@@ -1,83 +1,44 @@
...
+3. Executing a basic `SELECT 123 query` through the RDS Data API.
 
-DB Snapshots are currently not supported for MSSQL.
+LocalStack's RDS implementation also supports the [RDS Data API](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/data-api.html), which allows executing data queries against RDS clusters over a JSON/REST interface. 
 
-## End-to-end example (Postgres)
+### Create an RDS cluster
 
-The local RDS service also supports the [RDS Data API](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/data-api.html), which allows executing data queries against RDS clusters over a JSON/REST interface. 
+To create an RDS cluster, you can use the [`CreateDBCluster`](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBCluster.html) API. The following command creates a new cluster with the name `db1` and the engine `aurora-postgresql`. The cluster will be created with a single instance, which will be used as the master instance. The instance will be created with the default username `test` and password `test`.
```
steffyP, line 28: > The instance will be created with the default username `test` and password `test`
IMO we should not outline this here, for a few reasons. 

- The sample below actually sets a username + password (which is different from the default) and it might confuse people. 
- In general it's not possible to leave the fields blank for the real AWS, so I think we shouldn't promote non-parity behavior.
- For some databases (MySQL + MSSQL) we spin up the official images which will not work with the defaults (for MySQL this just changed recently, and we rely on the user setting a valid user-password combination).

content/en/user-guide/aws/rds/index.md
```diff
@@ -98,17 +65,41 @@ $ cat << 'EOF' > mycreds.json
...
     "ARN": "arn:aws:secretsmanager:us-east-1:000000000000:secret:dbpass-cfnAX",
     "Name": "dbpass",
     "VersionId": "fffa1f4a-2381-4a2b-a977-4869d59a16c0"
 }
+```
+
+### Execute a query
+
+To execute a query, you can use the [`ExecuteStatement`](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ExecuteStatement.html) API. The following command executes a query against the database. The query returns the value `123`.
+
```
steffyP, line 91:
```suggestion
Make sure to replace the `secret-arn` with the arn from the secret you just created in the previous step. 
Also check that the `resource-arn` matches the cluster-arn that you have created before.
```

content/en/user-guide/aws/rds/index.md
```diff
@@ -137,69 +128,152 @@ $ awslocal rds-data execute-statement --database test --resource-arn arn:aws:rds
...
+    --engine postgres \
+    --db-name $DB_NAME \
+    --enable-iam-database-authentication \
+    --db-instance-class db.t3.small
+{{< / command >}}
+
+### Connect to the database
 
-$ awslocal rds create-db-instance --master-username $MASTER_USER --master-user-password $MASTER_PW --db-instance-identifier mydb --engine postgres --db-name $DB_NAME --enable-iam-database-authentication --db-instance-class db.t3.small
+You can retrieve the hostname and port of your created instance either from the preceding output or by using the [`DescribeDbInstances`](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBInstances.html) API. Run the following command to connect to the database:
```
steffyP, line 219:
```suggestion
You can retrieve the hostname and port of your created instance either from the preceding output or by using the [`DescribeDbInstances`](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBInstances.html) API. Run the following command to retrieve the host and port of the instance:
```

content/en/user-guide/aws/rds/index.md
```diff
@@ -137,69 +128,152 @@ $ awslocal rds-data execute-statement --database test --resource-arn arn:aws:rds
...
+
+- Creating a global database will result in the generation of a single local database. All clusters and instances associated with the global database will share a common endpoint.
+- It's important to note that clusters removed from a global database lose their ability to function as standalone clusters, differing from their intended behavior on AWS.
+- At present, the capability for persistence within global databases is not available.
+
+## Resource Browser
+
+The LocalStack Web Application provides a Resource Browser for managing RDS instances and clusters. You can access the Resource Browser by opening the LocalStack Web Application in your browser, navigating to the **Resources** section, and then clicking on **RDS** under the **Database** section.
+
+<img src="rds-resource-browser.png" alt="RDS Resource Browser" title="RDS Resource Browser" width="900" />
```
steffyP, line 259: For me the image is not showing in the preview (only showing the "alt" text):
https://localstack-docs-preview-pr-764.surge.sh/user-guide/aws/rds/#resource-browser. In the console it states `Image corrupt or truncated`. 
Could it have something to do with the preview? Because on the same deployment I see this error for the cloudwatch docs as well, but on our official, released docs it works just fine 🤔

steffyP (approved):
Thanks for tackling this @HarshCasper 🙏 
I added some comments, please check if we still want to rephrase those parts 🙂

content/en/user-guide/aws/rds/index.md
HarshCasper: Works fine on my end

image

content/en/user-guide/aws/rds/index.md
HarshCasper: Its just for an intro to AWS RDS, not LocalStack's implementation :)

PR #770, opened 2023-08-18
add docs for config

HarshCasper:


thrau (approved):
this was really well prepared @HarshCasper, thank you!

i just
* put every sentence in a line
* in the guide, removed the need to create a json file on disk
* added a limitations section saying that the service is only mocked

PR #775, opened 2023-08-23
add docs for efs

HarshCasper:


content/en/user-guide/aws/elastic-file-system/index.md
```diff
@@ -0,0 +1,104 @@
...
+---
+
+## Introduction
+
+EFS (Elastic File System) is a fully managed file storage service provided by Amazon Web Services (AWS). EFS offers scalable and shared file storage that can be accessed by multiple EC2 instances and on-premises servers simultaneously. EFS utilizes the Network File System protocol to allow it to be used as a data source for various applications and workloads.
+
+LocalStack supports EFS via the Pro/Team offering, allowing you to use the EFS APIs in your local environment to create local file systems, lifecycle configurations, and file system policies. The supported APIs are available on our [API coverage page](https://docs.localstack.cloud/references/coverage/coverage_efs/), which provides information on the extent of EFS's integration with LocalStack.
+
+{{< alert title="Note" >}}
+LocalStack's EFS implementation is limited and lacks support for functionalities like creating mount targets, configuring access points, and generating tags. LocalStack uses Moto to emulate the EFS APIs, and efforts are underway to incorporate support for these features in upcoming updates.
```
bentsku, line 15: nit: I don't think we use moto for EFS, at least I don' see it in `providers.py` in -ext.
In the coverage page we say `Emulated` for EFS, but I think it really is only mocked? I don't know how you feel about it, at least as long as we don't have mount target or access point support?

bentsku (commented):
Thanks for adding this! I'm going to add more AWS validated tests to EFS so I've started reading about it too. Just one comment about moto usage, which I don't think we do, but the rest is great! Thanks a lot!

content/en/user-guide/aws/elastic-file-system/index.md
cabeaulac: We have some mount target support. @whummer did a quick PoC for it recently.

bentsku: Is it really using EFS mount targets, or are we directly using Docker to mount a local folder into the container?

bentsku: I just saw that a PR got merged for EFS 😄 I'll have a look at the changes!

bentsku: We actually now use Moto, maybe you were already in the loop, sorry for the comment then 😁

PR #811, opened 2023-09-06
Fixed "disable copy" feature bug

webdev51:
This PR fixes a reported bug (screenshot attached below) in the newly created "disable-copy" feature as it was acting weird with new lines.

image


whummer (approved):
Awesome, thanks for the quick fix @webdev51 🙌

PR #817, opened 2023-09-11
refresh Glue docs, provide dev hub app shortcut

alexrashed:
This PR contains the following changes:
- Refresh the Glue service docs according to our new guidelines.
- Make sure every single sentence is in its own line in the markdown file.
- Restructure the Developer Hub code and provide a shortcode which renders all developer hub samples which integrate with a specific service.
  - [Preview Deployment](https://localstack-docs-preview-pr-817.surge.sh/user-guide/aws/glue/)
  - The result looks like this:
    image
  - This could be very easily integrated for other services (with a single line of code).
  - I think this is a nice promotion of the developer hub apps directly on the docs page of the respective services.
  - This is where I would like to get input from @HarshCasper @tinyg210 @lukqw if this makes sense or if I should revert the changes.

### DIFF
--- a/layouts/partials/developer-hub/application.html
+++ b/layouts/partials/developer-hub/application.html
@@ -6,7 +6,7 @@
 {{$cloudpod := cond (eq (.cloudPods) true) " cloudpod-true" "" }}
 {{$className := print $services $deployment $platform $level $cloudpod}}
 <div
-    class='entry application-card filters-card col-xs-12 col-sm-12 col-md-4{{$className}}'
+    class='entry application-card filters-card col-xs-12 col-sm-6 col-lg-4{{$className}}'
     data-services='{{cond (ne (len .services) 0) (delimit .services ",") ""}}'
     data-deployments='{{cond (ne (len .deployment) 0) (delimit .deployment ",") ""}}'
     data-platforms='{{cond (ne (len .platform) 0) (delimit .platform ",") ""}}'


lukqw (approved):
Cool addition regarding the dev hub apps 🚀 
We could think about introducing the same for our tutorials in the future.

alexrashed:
Thanks for the input!
Based on your feedback (and with @lukqw's help with adjusting the styling), I now also extracted the tutorial partial and created a `tutorials` shortcode.
The result looks as follows:
image

layouts/partials/developer-hub/application.html
```diff
@@ -0,0 +1,51 @@
+{{$servicesData := site.Data.developerhub.services }}
+{{$services := cond (ne (len .services) 0)  (printf " service-%s" (delimit .services " service-")) "" }}
+{{$deployment := cond (ne (len .deployment) 0)  (printf " deployment-%s" (delimit .deployment " deployment-")) "" }}
+{{$platform := cond (ne (len .platform) 0)  (printf " platform-%s" (delimit .platform " platform-")) "" }}
+{{$level := cond (ne (len .complexity) 0)  (printf " level-%s" (delimit .complexity " level-")) "" }}
+{{$cloudpod := cond (eq (.cloudPods) true) " cloudpod-true" "" }}
+{{$className := print $services $deployment $platform $level $cloudpod}}
+<div
+    class='entry application-card filters-card col-xs-12 col-sm-12 col-md-4{{$className}}'
```
lukqw, line 9: This should be changed to
```
    class='entry application-card filters-card col-xs-12 col-sm-6 col-lg-4{{$className}}'
```

That way it's consistent with the width of tutorial cards across all screen sizes.

Otherwise it looks like this on smaller screens:

Screenshot 2023-09-12 at 12 21 26

lukqw (approved):
Awesome job on adding the tutorials as well 🚀

layouts/partials/developer-hub/application.html
alexrashed: awesome, thanks for the suggestion!

PR #804, opened 2023-09-03
Add tutorial on using Crossplane AWS provider with LocalStack

whummer:
Add tutorial on using [Crossplane](https://crossplane.io/) AWS provider with LocalStack. Motivated by a recent user request in the Discuss forum: https://discuss.localstack.cloud/t/crossplane-integration/507 (addresses point (2) in my response in the thread)

As a follow-up, it would be great if we can update the overview figure in our [integrations page](https://docs.localstack.cloud/user-guide/integrations/) - it is a bit outdated at this point, and a couple more integrations have been added in the meantime. @HarshCasper maybe we can work with @lazarvu on that - we also have a figure in our customer deck that we can build upon. 👍 

---
@HarshCasper I was trying to follow your advice on not using multi-line `<command>` listings that also contain the command output (to prevent users from copying the entire text when using the "copy" button), but tbh it was a bit tricky, as always separating the command and output into separate listings blows up the length of the text, and I kind of like the terse syntax with commands+output combined. Maybe we can introduce a new tag that would render commands, but with the "copy" button disabled? 🤔  To be discussed - happy to hear your thoughts as well..

PR #841, opened 2023-10-03
Split EKS commands and outputs for better copy and paste

simonrw:
# Motivation

When starting an EKS cluster, I was copying the commands from the docs but the output was copied along with the commands.

# Changes

* Split the command and output into separate code blocks

### DIFF
--- a/content/en/user-guide/aws/elastic-kubernetes-service/index.md
+++ b/content/en/user-guide/aws/elastic-kubernetes-service/index.md
@@ -57,27 +57,20 @@ You can use the `docker` CLI to check that some containers have been created:
 
 {{< command >}}
 $ docker ps
-{{< / command >}}
-
-You can see an output similar to the following:
-
-```
+<disable-copy>
 CONTAINER ID   IMAGE                          COMMAND                  CREATED          STATUS          PORTS                                           NAMES
 ...
 b335f7f089e4   rancher/k3d-proxy:5.0.1-rc.1   "/bin/sh -c nginx-pr…"   1 minute ago   Up 1 minute   0.0.0.0:8081->80/tcp, 0.0.0.0:44959->6443/tcp   k3d-cluster1-serverlb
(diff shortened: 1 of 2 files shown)


alexrashed (commented):
Great that you are changing the docs right when coming across issues!
For some of the changes in the PR it might make sense to use the new feature @webdev51 implemented: https://github.com/localstack/docs/pull/804#discussion_r1315160963
Wit that you can disable the copying for certain sections within the command, instead of separating command and output all the time (which sometimes makes sense, but sometimes disrupts the flow).

simonrw:
> Great that you are changing the docs right when coming across issues! For some of the changes in the PR it might make sense to use the new feature @webdev51 implemented: [#804 (comment)](https://github.com/localstack/docs/pull/804#discussion_r1315160963) Wit that you can disable the copying for certain sections within the command, instead of separating command and output all the time (which sometimes makes sense, but sometimes disrupts the flow).

Thanks @alexrashed for the suggestion. I definitely prefer the output as rendered, however the copied command contains a newline at the end. An alternative is to do something like

```markdown
{{< command >}}
$ pwd<disable-copy>
/
</disable-copy>
{{< / command >}}
```

which is a little ugly, but in general I prefer to not break up the reading flow as you say.

alexrashed:
Ah, interesting. A newline at the end really doesn't seem to be too nice when copy-pasting the commands.
Maybe we could fix the command shortcode to suppress the last newline?
Should be fixed somewhere around this line:
https://github.com/localstack/docs/blob/main/layouts/shortcodes/command.html#L6

simonrw:
> Ah, interesting. A newline at the end really doesn't seem to be too nice when copy-pasting the commands. Maybe we could fix the command shortcode to suppress the last newline?

[I have implemented this by trimming the whitespace from the copied text before sending to the clipboard](https://github.com/localstack/docs/pull/841/commits/d634eec251eeb8770c0d9a8d495c57f8d3cd0b14).

alexrashed (approved):
Awesome! 💯

PR #847, opened 2023-10-06
update elb docs 

sannya-singal:
While running the elb sample instructions figured out that in some cases we might to give the target id while registering the target as the `Gateway` address of the docker container.


macnev2013 (approved):
Thanks for updating the docs. LGTM 🚀

PR #848, opened 2023-10-07
add docs for import/export feature

HarshCasper:


lukqw (approved):
This looks great!
Thanks for adding documentation around this 🚀

giograno:
Thanks @HarshCasper for this! 💯 . We could probably extend the docs once [this](https://github.com/localstack/localstack-ext/pull/2167) is merged. 

Side question: do we need to start using different terms for the "local pods" and the ones stored in the remotes?

PR #851, opened 2023-10-10
Clarify Lambda layers pro feature

joe4dev:
The Lambda documentation does not clarify that "Lambda layers" is a Pro feature.

We should be consistent with the coverage docs marking "Lambda layers" as a Pro feature: https://docs.localstack.cloud/user-guide/aws/feature-coverage/
Screenshot 2023-10-10 at 16 05 06

## Changes

* Add `(Pro)` to the Lambda layers title to be consistent with the coverage docs
* Clarify the behavior by explaining that CRUD still works in LocalStack Community but usage is Pro
* Fix phrasing (layers cannot be invoked)


Pive01 (approved):
LGTM! Thanks for the clarification!

PR #861, opened 2023-10-15
improve terraform docs

HarshCasper:


whummer (commented):
Just two tiny orthographic nitpicks. But would like to solicit feedback from @lakkeger here as well. 👍

PR #866, opened 2023-10-17
reformat transparent endpoint injection docs

HarshCasper:


simonrw:
@HarshCasper there are quite a few build errors showing up. I'm happy to review once these errors have been fixed.

joe4dev:
oh, I just spotted this one after doing the migration for removing the legacy lambda provider and updating some content to adopt `AWS_ENDPOINT_URL` in https://github.com/localstack/docs/pull/899

Integrating the sub-page makes sense now that we only have one unless we treat the new `AWS_ENDPOINT_URL` as a separate way of transparent endpoint injection 🤔

simonrw:
@HarshCasper @joe4dev is this PR still valid given #941?

PR #877, opened 2023-10-25
SQS: Add new endpoint strategy

viren-nadkarni:
This PR updates user docs with details about the new SQS endpoint strategy 

Depends on https://github.com/localstack/localstack-artifacts/pull/20


alexrashed (changes requested):
On the one hand, this PR depends on https://github.com/localstack/localstack-artifacts/pull/20 (because before that the endpoints cannot be properly used with HTTPS enabled), and the advertisement of the new endpoint strategy also depends on how you decide to rename them for v3:
https://github.com/localstack/localstack/pull/9454#pullrequestreview-1698147035
Are you going to rename them from `domain` to `legacy`, and then renaming `standard` to `domain`. Because if so, I would not add this to the docs like this right now, but wait for the release and describe the new way (otherwise users who switch to `standard` now, will experience a breaking change with 3.0).

viren-nadkarni:
> Are you going to rename them from `domain` to `legacy`, and then renaming `standard` to `domain`. Because if so, I would not add this to the docs like this right now, but wait for the release and describe the new way (otherwise users who switch to `standard` now, will experience a breaking change with 3.0).

I'm against renaming the existing strategies to avoid confusion among users. It is possible that we might also break user workflows in doing so.

alexrashed:
That's a good argument! But then let's still wait for the cert to be updated before we advertise the new endpoint strategy...

alexrashed (approved):

PR #882, opened 2023-10-31
fix scheduled strategy docs

giograno:
This PR removes a duplicate entry for the `SCHEDULED` persistence strategy.
Thanks to @whummer for noticing it!


content/en/references/persistence-mechanism.md
```diff
@@ -61,8 +61,6 @@ You can select a particular save strategy by setting `SNAPSHOT_SAVE_STRATEGY=<st
 * **`SCHEDULED`**: (**default**): saves at regular intervals the state of all the services that have been modified since the last snapshot.
```
whummer, line 61:
```suggestion
* **`SCHEDULED`** (**default**): saves at regular intervals the state of all the services that have been modified since the last snapshot.
```

whummer (approved):
Thanks for the quick fix! 👍

PR #895, opened 2023-11-07
Remove DEFAULT_REGION and USE_SINGLE_REGION

viren-nadkarni:
This PR updates the docs to reflect the changes made in https://github.com/localstack/localstack/pull/9375 and https://github.com/localstack/localstack/pull/9374


alexrashed (approved):
Thanks for cleaning up the newly removed (but quite old) config variables! 🧹

PR #897, opened 2023-11-08
Remove EC2 local daemon

viren-nadkarni:
The EC2 local daemon will be removed in the upcoming v3 release.

This removes the outdated portions of user docs. It also highlights some other important lines in alert boxes so that they stand out.


alexrashed (approved):
Thanks for adjusting the docs with the upcoming changes! 🚀 🧹

PR #898, opened 2023-11-08
IoT endpoint are now fully qualified

viren-nadkarni:
With LocalStack v3, IoT MQTT broker addresses contain the account ID and region.


alexrashed (approved):
LGTM! 🧹

PR #899, opened 2023-11-08
Remove lambda legacy options

joe4dev:
## Changes

* Remove configuration options for the old Lambda provider
* Remove patched SDKs under transparent endpoint injection only relevant for the old provider
* Update transparent endpoint injection overview section

## Notes

The migration guide is still kept in the Lambda page for a while to help people migrating that still used the legacy provider.


dominikschubert (approved):
LGTM! Thanks for updating the lambda docs :rocket:

PR #892, opened 2023-11-06
Major: rename default container name to localstack-main

simonrw:
# Motivation

Hyphens are not allowed in URLs. When addressing the LocalStack container over a docker network, the container name resolves to the IP address of the container. This is a problem in particular for boto3 which refuses to connect.

# Changes

This PR updates usages of `localstack_main` to `localstack-main`.

PR #905, opened 2023-11-09
Fix CI broken link checks

simonrw:
# Motivation

It seems merging #892 broke the broken link checker. This may be because the config file previously had an incorrect format, and was therefore invalid JSON, however it is confusing that the pipeline did not fail at any point.

Regardless, some broken links were found:

* the use of `local.localstack.cloud` (rather than `localhost.localstack.cloud`)
* links to removed neptune API reference

# Changes

This PR fixes the two points listed above, and fixes the broken link checker.


joe4dev (approved):
LGTM 🚀 
Thanks for jumping on that.

PR #906, opened 2023-11-09
Partition old configuration into deprecation/legacy

dominikschubert:
Introduces a new partitioning of older configuration values to unify how we communicate their deprecation status.

1. `Deprecated` ... Still supported in latest version but should not be used anymore. Will be removed in one of the next major releases.
2. `Legacy` ... Has been removed in a previous major release. Does **NOT** affect the latest version anymore.

## Follow-up

We should probably also cover variables we've just removed from the list in the past.

This isn't super urgent but we should make sure we sync the deprecation status here with the ones we're tracking in localstack.

### DIFF
--- a/content/en/references/configuration.md
+++ b/content/en/references/configuration.md
@@ -314,8 +314,8 @@ These configurations are deprecated and will be removed in the upcoming major ve
 
 | Variable | Example Values | Description |
 | - | - | - |
-| `SKIP_INFRA_DOWNLOADS` | | **Deprecated.** Whether to skip downloading additional infrastructure components (e.g., specific Elasticsearch versions) |
 | `BIGDATA_MONO_CONTAINER` | `0`\|`1` (default) | **Deprecated.** Whether to spin Big Data services inside the LocalStack main container. Glue jobs breaks when using `BIGDATA_MONO_CONTAINER=0`. | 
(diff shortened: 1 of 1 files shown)


joe4dev (approved):
Great initiative 🚀 . I like the separation and metadata about when it was removed.

Improved configuration handling and communication is something we should tackle 📈 

Two discussion points:

1. How long should we keep the legacy variables at all? one extra major release, forever?
2. How do we go about adding the deprecation flag inline (i.e., within the subsection) vs. globally?

For Lambda, we still have the [migration guide](https://docs.localstack.cloud/user-guide/aws/lambda/#migrating-to-lambda-v2) but I'm also fine with re-adding them for configuration discoverability at one page, if labelled clearly.

dominikschubert:
> The most urgent thing is probably to synchronize the list with deprecations.py and what we actually remove in 3.0 🧙

I wouldn't call it urgent, but yeah definitely want to have this synced until next week!
Basically I'd like to introduce the same terminology there as well.

dominikschubert:
> How long should we keep the legacy variables at all? one extra major release, forever?
Yes that was our thinking. At least forever mentioning that they once existed.

> How do we go about adding the deprecation flag inline (i.e., within the subsection) vs. globally?
It's probably not needed anymore to have it inline, I'll remove this in a follow-up when adding more items

PR #916, opened 2023-11-15
Update Pulumi.dev.yaml content

MarcelStranak:
The list provided in the documentation is outdated, and when you try to use `pulumi up`, it ends up with an error.

```
error: pulumi:providers:aws resource 'default_6_8_0' has a problem: could not validate provider configuration: Invalid or unknown key. Check pulumi config get aws:endpoints.
```

### DIFF
--- a/content/en/user-guide/integrations/pulumi/index.md
+++ b/content/en/user-guide/integrations/pulumi/index.md
@@ -70,313 +70,104 @@ Now edit your stack configuration `Pulumi.dev.yaml` as follows:
 
 ```yaml
 config:
+  aws:region: us-east-1
   aws:accessKey: test
   aws:secretKey: test
-  aws:s3ForcePathStyle: 'true'
-  aws:skipCredentialsValidation: 'true'
-  aws:skipRequestingAccountId: 'true'
+  aws:skipCredentialsValidation: "true"
+  aws:skipRequestingAccountId: "true"
   aws:endpoints:
-    - accessanalyzer: http://localhost:4566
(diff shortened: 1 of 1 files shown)


content/en/user-guide/integrations/pulumi/index.md
```diff
@@ -70,313 +70,104 @@ Now edit your stack configuration `Pulumi.dev.yaml` as follows:
...
 config:
+  aws:region: us-east-1
   aws:accessKey: test
   aws:secretKey: test
-  aws:s3ForcePathStyle: 'true'
-  aws:skipCredentialsValidation: 'true'
-  aws:skipRequestingAccountId: 'true'
+  aws:skipCredentialsValidation: "true"
+  aws:skipRequestingAccountId: "true"
   aws:endpoints:
```
lakkeger, line 78: I'm not sure it's a good practice to provide a drop-in project file for the users with all the available endpoints without a disclaimer or note, because the list can change by every version of the package.
Perhaps just giving a couple of examples and point them to the actual package docs would be a better idea.
Docs: https://www.pulumi.com/registry/packages/aws/api-docs/provider/#providerendpoint

MarcelStranak: Hard to say for me. I prefer complete docs. We do have some similar docs for [Terraform](https://docs.localstack.cloud/user-guide/integrations/terraform/#endpoint-configuration), but they also look outdated.

@HarshCasper can you please share your input on this?

PR #933, opened 2023-11-23
Add new Lambda config LAMBDA_LIMITS_CODE_SIZE_ZIPPED

joe4dev:
Follow up of https://github.com/localstack/localstack/pull/9697

## Motivation

As detected in snapshot updates and implemented in https://github.com/localstack/localstack/pull/9697, AWS changed the behavior related to Lambda limits. Therefore, we should pro-actively document the option `LAMBDA_LIMITS_CODE_SIZE_ZIPPED` because everyone who used `LAMBDA_LIMITS_CREATE_FUNCTION_REQUEST_SIZE`, likey needs `LAMBDA_LIMITS_CODE_SIZE_ZIPPED` instead or in addition.

Further clarifications in the localstack PR, including this comment replying to @dfangl question https://github.com/localstack/localstack/pull/9697#discussion_r1402046114
> My guess: The previous error message was too confusing. People uploaded large zips and got a message about request size:
Request must be smaller than 70167211 bytes for the CreateFunction operation
>
> Now they check the zip (if relevant) before checking the request size limit leading to a clearer message:
Zipped size must be smaller than 52428800 bytes

## Changes

* Change the default of `LAMBDA_LIMITS_CREATE_FUNCTION_REQUEST_SIZE` following an AWS parity update
* Add the configuration `LAMBDA_LIMITS_CODE_SIZE_ZIPPED` to the documentation because that it now checked first after a recent AWS update. Therefore, users who currently adjusted `LAMBDA_LIMITS_CREATE_FUNCTION_REQUEST_SIZE`, likely need `LAMBDA_LIMITS_CODE_SIZE_ZIPPED` as well


dfangl (approved):
Looks good, thanks for documenting!

PR #935, opened 2023-11-24
improve documentdb docs

steffyP:
Improving the sample, and removing the `LOCALSTACK_HOST` usage.

### DIFF
--- a/content/en/user-guide/aws/docdb/index.md
+++ b/content/en/user-guide/aws/docdb/index.md
@@ -12,8 +12,6 @@ can be accessed from localhost.
 
 When defining a port to access the container, an available port on the host machine will be selected, that means there is no pre-defined port range by default.
 
-Using the flag `DOCDB_PROXY_CONTAINER=1` the default behavior changes and the container will be started as proxied container. Meaning a port from the [pre-defined port]({{< ref "/references/external-ports" >}}) range will be chosen, and when using lambda, you can use `LOCALSTACK_HOSTNAME` to connect to the instance.
-
 {{< alert title="Information" color="success">}}
 MongoDB is a popular open-source, document-oriented NoSQL database that provides high scalability, flexibility, and performance for modern application development.
 It belongs to the family of
@@ -174,9 +172,9 @@ databases.
 
 {{< command >}}
 
-$ mongosh mongodb://localhost.localstack.cloud:39045
+$ mongosh mongodb://localhost:39045
 Current Mongosh Log ID:    64a70b795697bcd4865e1b9a
-Connecting to:        mongodb://localhost.localstack.cloud:
+Connecting to:        mongodb://localhost:
 39045/?directConnection=true&serverSelectionTimeoutMS=2000&appName=mongosh+1.10.1
 Using MongoDB:        6.0.7
 Using Mongosh:        1.10.1
@@ -195,9 +193,9 @@ which is the cluster port that appears in the aforementioned description.
 To work with a specific database, the command is:
 
 {{< command >}}
-$ mongosh mongodb://localhost.localstack.cloud:39045/test-company
+$ mongosh mongodb://localhost:39045/test-company
 Current Mongosh Log ID:    64a71916fae7fdeeb8b43a73
-Connecting to:        mongodb://localhost.localstack.cloud:
+Connecting to:        mongodb://localhost:
 39045/test-company?directConnection=true&serverSelectionTimeoutMS=2000&appName=mongosh+1.10.1
 Using MongoDB:        6.0.7
 Using Mongosh:        1.10.1
@@ -232,7 +230,9 @@ the [MongoDB documentation](https://www.mongodb.com/docs/).
 ### Connect to DocumentDB using Node.js Lambda
 
 {{< alert title="Important" color="success">}}
-You need to set `DOCDB_PROXY_CONTAINER=1` when starting LocalStack to be able to use the `localhost.localstack.cloud` in lambda to connect to the instance.
+You need to set `DOCDB_PROXY_CONTAINER=1` when starting LocalStack to be able to use the returned `Endpoint`, which will be correctly resolved automatically.
+
+The flag `DOCDB_PROXY_CONTAINER=1` changes the default behavior and the container will be started as proxied container. Meaning a port from the [pre-defined port]({{< ref "/references/external-ports" >}}) range will be chosen, and when using lambda, you can use `localhost.localstack.cloud` to connect to the instance.
 {{< /alert>}}
 
 In this sample we will use Node.js lambda function to connect to a DocumentDB. 
@@ -262,7 +262,7 @@ In your terminal run:
 $ mkdir resources
 $ cd resources
 $ mkdir node_modules
-$ npm install mongodb
+$ npm install mongodb@6.3.0
 {{< /command >}}
 
 Next, copy the following code into a new file named `index.js` in the `resources` folder:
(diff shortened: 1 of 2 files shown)


content/en/user-guide/aws/docdb/index.md
```diff
@@ -231,16 +232,125 @@ the [MongoDB documentation](https://www.mongodb.com/docs/).
...
+
+The `outfile` contains the returned value, e.g.:
+```yaml
+{"statusCode":200,"body":"{\"_id\":\"6560a21ca7771a02ef128c72\",\"key\":\"value\"}"}
+````
+
+#### Use Secret To Connect to DocDB
+
+The best-practise for accessing databases is by using secrets. 
+Secrets follow a [well-defined pattern](https://docs.aws.amazon.com/secretsmanager/latest/userguide/create_database_secret.html).
```
joe4dev, line 350: Great idea to link to the official AWS docs 👍

content/en/user-guide/aws/docdb/index.md
```diff
@@ -231,16 +232,125 @@ the [MongoDB documentation](https://www.mongodb.com/docs/).
...
-* A helper function makes sure the username and password are encoded correctly for the connection URI.
-* We run some sample commands to verify the connection works.
+In this sample we will use Node.js lambda function to connect to a DocumentDB. 
+For the mongo-db connection we will use the `mongodb` lib.
+Please note, that this sample is only for demo purpose, e.g., we will set the credentials as environment variables to the lambda function.
+
+In a best-practise sample you would use a secret instead.
+We included a snippet at the very end.
+
+#### Create the DocDB Cluster with a username and password
```
joe4dev, line 245: I think it is great to align the DB creation with the sample code snippet and make it more self-contained.

One concern: The current page structure gives the impression that after following the first parts of getting started, the `Nodejs` section shows how to connect to the same cluster (using the employees, customers theme).

content/en/user-guide/aws/docdb/index.md
```diff
@@ -231,16 +232,125 @@ the [MongoDB documentation](https://www.mongodb.com/docs/).
...
+
+const docdb_client = new RDS();
+
+const docdb_id = process.env.DOCDB_CLUSTER_ID;
+const pwd = process.env.DOCDB_SECRET;
+
+exports.handler = async (event) => {
+  try {
+    // Get endpoint details using rds/docdb client:
+    var cluster_result = await docdb_client.describeDBClusters({DBClusterIdentifier: docdb_id}).promise();
```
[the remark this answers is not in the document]
joe4dev: This probably comes from trying to avoid `jq` parsing in bash ...

content/en/user-guide/aws/docdb/index.md
```diff
@@ -231,16 +232,125 @@ the [MongoDB documentation](https://www.mongodb.com/docs/).
 ### Connect to DocumentDB using Node.js Lambda
```
joe4dev, line 232: It would be very helpful (for us) to link to the scenario test you mentioned here: https://github.com/localstack/localstack-ext/blob/master/tests/aws/scenario/rds_neptune_docdb/test_rds_neptune_docdb.py#L112
(ideally vice-versa)

Even if it's just an internal comment (not visible in the docs).

joe4dev (changes requested):
Thank you for improving these docs greatly @steffyP 👏👏👏 

My main concern: I think the example is too complex for a standalone getting-started guide without providing any working sample repository. I am worried that customers won't get the best value if it already takes us way too much time to set up. If we want to keep these docs updated in the future, we need to go through them step-by-step ...
Counterargument: Your improvements made this more feasible and keeping it in sync with the scenario test could be a first step (also helps for AWS parity).

I added some more comments in the review (just my thoughts, feel free to ignore them :)

content/en/user-guide/aws/docdb/index.md
```diff
@@ -231,16 +232,125 @@ the [MongoDB documentation](https://www.mongodb.com/docs/).
...
+    const dbname = "mydb";
+    // retryWrites is by default true, but not supported by AWS DocumentDB
+    const uri = `mongodb://${user}:${pwd}@${host}:${port}/?retryWrites=false`;
+    
+    // Connect to DocumentDB
+    const client = await MongoClient.connect(uri);
+    const db = client.db(dbname);
+
+    // Insert data
+    const collection = db.collection('your_collection');
```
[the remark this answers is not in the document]
steffyP: no

content/en/user-guide/aws/docdb/index.md
```diff
@@ -231,16 +232,125 @@ the [MongoDB documentation](https://www.mongodb.com/docs/).
 ### Connect to DocumentDB using Node.js Lambda
 
 {{< alert title="Important" color="success">}}
-You need to set `DOCDB_PROXY_CONTAINER=1` when starting LocalStack to be able to use the `LOCALSTACK_HOSTNAME` in lambda to connect to the instance.
+You need to set `DOCDB_PROXY_CONTAINER=1` when starting LocalStack to be able to use the `localhost.localstack.cloud` in lambda to connect to the instance.
```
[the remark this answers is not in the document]
steffyP: we already explain it at the very beginning of the docs:

> Using the flag DOCDB_PROXY_CONTAINER=1 the default behavior changes and the container will be started as proxied container. Meaning a port from the [pre-defined port](https://localstack-docs-preview-pr-935.surge.sh/references/external-ports/) range will be chosen, and when using lambda, you can use LOCALSTACK_HOSTNAME to connect to the instance.

what else would you add here? :)

content/en/user-guide/aws/docdb/index.md
steffyP: only if we don't move it 😅 
stupid question: is there something in hugo that we can use to add comments, that will not be rendered/showing up in the docs? 🤔

content/en/user-guide/aws/docdb/index.md
steffyP: one way or another we want to show both options: once without user/pw, and one with. 
IMO this way it's less confusing 🤷‍♀️

content/en/user-guide/aws/docdb/index.md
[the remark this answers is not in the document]
joe4dev: Yeah, that's exactly what I was looking for 👍 . It might be better to move it here because it is mostly relevant in that context 🤷 .

We should also update the reference to `LOCALSTACK_HOSTNAME`, also in the config docs https://docs.localstack.cloud/references/configuration/#documentdb

content/en/user-guide/aws/docdb/index.md
steffyP: found something

joe4dev: HTML or Go template comments should do it: https://gohugo.io/templates/introduction/#comments

content/en/user-guide/aws/docdb/index.md
```diff
@@ -231,16 +230,129 @@ the [MongoDB documentation](https://www.mongodb.com/docs/).
...
 {{< /alert>}}
 
-The code snippet is a Node.js lambda.
-In this sample we assume that we have a secret where the DocumentDB credentials are stored.
-* The secret arn is passed as env `SECRET_NAME`.
-* The Lambda reads the connection details and connects to the running DocumentDB, which runs in proxied docker container. 
-* As we run inside lambda, we change the host to the `LOCALSTACK_HOSTNAME` to connect to the container.
-* A helper function makes sure the username and password are encoded correctly for the connection URI.
-* We run some sample commands to verify the connection works.
+In this sample we will use Node.js lambda function to connect to a DocumentDB.
```
joe4dev, line 238: missing article

joe4dev (approved):
Tested on macOS against LocalStack `latest` (2023-11-24) 👍 

It works as a self-contained getting started guide and looks a big step cleaner 👏 👏 

LGTM ✔️

content/en/user-guide/aws/docdb/index.md
steffyP, line 238:
```suggestion
In this sample we will use a Node.js lambda function to connect to a DocumentDB. 
```

PR #941, opened 2023-11-28
Rework dns server and transparent endpoint injection

joe4dev:
## Motivation

The current DNS server page is badly outdated, misleading, and incomplete. With the increasing importance of DNS-based resolution in LocalStack networking, the DNS server page needs to be reworked.

## Changes

* Document missing config `DISABLE_TRANSPARENT_ENDPOINT_INJECTION`
* Restructure the DNS server page
  * Add more subheaders
  * Adjust order by de-emphasizing the system configuration because that is more of a special case. Information that is relevant for LocalStack container users is more important and should come first.
* Refine and extend the general introduction to this page
* Re-write system DNS config section with a general start and OS-specific subsections. macOS is done and tested.
* Generally clarify writing and warnings
* **UPDATE**: Move the DNS server page to a top-level tool and describe transparent endpoint injection more self-contained
* Ignore problematic URLs in link checker:
  * CloudFlare using bot detection returning 403 forbidden
  * GithHub boto3 link with line reference frequently flakes. Unfortunately, retries only happen on 429 (too many requests) according to the link [checker tool](https://github.com/tcort/markdown-link-check#api).
* **UPDATE**: Consistently use `"Transparent Endpoint Injection"` instead of transparent execution mode

## Preview

* Transparent Endpoint Injection: https://localstack-docs-preview-pr-941.surge.sh/user-guide/tools/transparent-endpoint-injection
* DNS Server: https://localstack-docs-preview-pr-941.surge.sh/user-guide/tools/dns-server

## Discussion

* The explanations around "transparent execution mode" / "transparent endpoint injection" could be improved. We used to have multiple implementations (legacy SDK replacement + DNS-based) and with the introduction of `AWS_ENDPOINT_URL`, this could be considered another implementation because it does transparently inject the endpoint. However, the configuration `DISABLE_TRANSPARENT_ENDPOINT_INJECTION` could be confusing if we do not clarify that this specific implementation refers to the DNS-based implementation.
* It is not 100% clear where transparent endpoint injection is explained. The page [Transparent Endpoint Injection](https://localstack-docs-preview-pr-941.surge.sh/user-guide/tools/transparent-endpoint-injection/) (preview) gives a short 1-sentence intro and then refers to the DNS server page for more details. Not yet ideal (but much better than before ;).
* The default for `DNS_SERVER` mentioned to be `8.8.8.8` can be confusing because, in the current implementation, the upstream DNS will be used as default (i.e., typically the system DNS). That is also the reason why we need to explicitly set an external DNS server when configuring LocalStack as system DNS because otherwise, we would create an infinite loop ;)

## TODO

- [x] Clarify default DNS server behavior
- [x] Fix CloudFlare link yielding 403 (probably checking user agent).
  - CloudFlare doesn't like link checkers an blocks them using some [advanced fingerprinting](https://www.zenrows.com/blog/cloudflare-403-forbidden-bypass) (user-agent worked on my machine). Ignoring the URL for now.

## Follow Up

- [ ] @simonrw @dfangl Update Linux instructions for system DNS configuration
- [ ] @simonrw Adjust docs for new `DNS_SKIP_LOCAL_RESOLUTION` https://github.com/localstack/localstack/pull/9692


simonrw:
The DNS server is still under the "transparent endpoint injection" section, is this a conscious decision? I think it would be better either at the top level in the user guide, or at least elevated to the "LocalStack tools" section, though it's not an external tool like the section naming suggests.

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
```diff
@@ -8,52 +8,88 @@ aliases:
...
 
 ```bash
-DNS_SERVER=1.1.1.1
+DNS_LOCAL_NAME_PATTERNS='.*(ecr|lambda).*.amazonaws.com'
 ```
 
-Per default, LocalStack uses the Google DNS resolver at `8.8.8.8`.
+Using this configuration, the LocalStack DNS server only resolves ECR and Lambda domains to LocalStack, and the rest will be resolved via the `DNS_SERVER` (i.e., the real DNS entry by default).
+This can be used for hybrid setups, where certain API calls (e.g., ECR, Lambda) target LocalStack, whereas other services will target real AWS.
+The regex pattern follows Python flavored-regex and can be tested at [regex101.com](https://regex101.com/r/OzIsQa/1).
```
simonrw, line 78: **This is why we need to rename this variable!!!!!**.

Any queries matching these regexes are _ignored_ by the LocalStack DNS server, so with TEI on anything _not_ matching these regexes is resolved by LocalStack, and `*.ecr.*.amazonaws.com` and `*.lambda.*.amazonaws.com` are forwarded to the upstream resolver.

Here we split the regex by whitespace and add each entry to the skip list in the resolver:
https://github.com/localstack/localstack/blob/4b7482d22e81bfff8c43f8b74eaeadb60bc6debf/localstack/dns/server.py#L831-L833


and here we skip local resolution if the domain matches:
https://github.com/localstack/localstack/blob/0bb7ce612be52a65c218377cdb04a14a215c6ba1/localstack/dns/server.py#L398-L409

```suggestion
Using this configuration, the LocalStack DNS server resolves all AWS domains to LocalStack _except_ ECR and Lambda domains which will be resolved via the `DNS_SERVER` (i.e., the real DNS entry by default).
This can be used for hybrid setups, where certain API calls (e.g., ECR, Lambda) target AWS, whereas other services will target LocalStack.
The regex pattern follows Python flavored-regex and can be tested at [regex101.com](https://regex101.com/r/OzIsQa/1).
```

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
```diff
@@ -8,52 +8,88 @@ aliases:
...
+Use this configuration with caution because we generally do not recommend connecting to real AWS from within LocalStack. 
+{{< /alert >}}
+
 
 ## Self-signed certificates
 
-When you configure transparent execution mode using DNS, you may still have to configure your application's AWS SDK to accept self-signed certificates.
-This is a technical limitation caused by the SSL certificate validation mechanism, due to the fact that we are repointing AWS domain names (e.g., `*.amazonaws.com`) to `localhost`. For example, the following command will fail with an SSL error:
+In the transparent execution mode using DNS in LocalStack Pro, you may still have to configure your application's AWS SDK to accept self-signed certificates.
+This is a technical limitation caused by the SSL certificate validation mechanism, due to the fact that we are repointing AWS domain names (e.g., `*.amazonaws.com`) to `localhost`.
```
simonrw, line 91:
```suggestion
This is a technical limitation caused by the SSL certificate validation mechanism, due to the fact that we are repointing AWS domain names (e.g., `*.amazonaws.com`) to `localhost.localstack.cloud`.
```

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
```diff
@@ -8,52 +8,88 @@ aliases:
...
+
+[The regex101 link is maintained by Joel Scheuner (requires linking to GitHub or Google account).
+It redirects to the main page if the saved example would not work.]: #
+
+{{< alert title="Warning" color="warning">}}
+Use this configuration with caution because we generally do not recommend connecting to real AWS from within LocalStack. 
+{{< /alert >}}
+
 
 ## Self-signed certificates
```
simonrw, line 88: I think this step could remain with the TEI docs, since it is not specific to the DNS server but rather TEI.

joe4dev:
> The DNS server page is still under the "transparent endpoint injection" section, is this a conscious decision? I think it would be better either at the top level in the user guide, or at least elevated to the "LocalStack tools" section, though it's not an external tool like the section naming suggests.

@HarshCasper suggested the same in https://github.com/localstack/docs/pull/866 and that makes sense if it would remain the only subsection


I foresee that we are going to have 3 scenarios on how to connect to LocalStack:
1. [Community] Manual endpoint configuration: Specify the `endpoint_url` using `AWS_ENDPOINT_URL` for AWS clients that do not yet support it natively (all default Lambda runtimes ATM: https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html). This is not really a tool but rather something we should document for each (major) runtime (currently missing). Maybe in the Lambda docs?
2. [Community] Auto `AWS_ENDPOINT_URL` detection: Newer SDKs automatically detect `AWS_ENDPOINT_URL` and no code changes are necessary. This is likely going to be the default option in the future. This is also not really a tool and does not fit here nicely.
3. [Pro] LocalStack Transparent Endpoint injection: The current way of DNS-based seamless connectivity to LocalStack for Pro users is enabled by default in Lambda (i.e., opt-out). This is a LocalStack tool and fits here.

Given that only the LS transparent endpoint injection is a real (internal) "tool", it might best to only cover that one here (including `DISABLE_TRANSPARENT_ENDPOINT_INJECTION`) and move the "DNS server" somewhere else.

> I think it would be better either at the top level in the user guide, or at least elevated to the "LocalStack tools" section, though it's not an external tool like the section naming suggests.

Transparent endpoint injection and Lambda are also not really "external" tools (I think the term is used broadly for extra features beyond AWS emulation). I agree/suggest moving the "DNS server" to the top-level Tools section for now. What do you think?

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
```diff
@@ -82,60 +120,114 @@ Disabling SSL validation may have undesired side effects and security implicatio
 Make sure to use this only for local testing, and never in production.
 {{< /alert >}}
 
+
+## Customizing internal endpoint resolution
+
+The DNS name `localhost.localstack.cloud` (and any subdomains like `mybucket.s3.localhost.localstack.cloud`) is used internally in LocalStack to route requests, e.g., between a Lambda container and the LocalStack APIs.
+
+Please refer to the steps in the [Route53 docs]({{< ref "route53" >}}) for more details on how the internal DNS name can be customized.
```
simonrw, line 128: I'm not sure I follow what this section is documenting. It is talking about how we use `localhost.localstack.cloud` internally to communicate between LocalStack components, but then it mentions how this value can be customised by Route53. I would suggest that mentioning Route53 is _additive_, but also that `LOCALSTACK_HOST` can be used to customise the actual domain used.

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
```diff
@@ -82,60 +120,114 @@ Disabling SSL validation may have undesired side effects and security implicatio
...
 
-In order to use transparent execution mode, the system needs to be configured to use the predefined DNS server.
-This is necessary if you want to test code running directly on your system against LocalStack, instead of AWS.
-The configuration depends on the operating system.
+If you wish to use the DNS server on your host system, you need to expose the LocalStack DNS server and configure your operating system.
+This is necessary if you want to test unmodified application code directly on your system against LocalStack and cannot configure the endpoint URL.
 
 {{< alert title="Warning" color="warning">}}
 Please be careful when changing the network configuration on your system, as this may have undesired side effects.
+Restore the default configuration after testing.
```
simonrw, line 177:
```suggestion
Remember to restore the default configuration after testing.
```

perhaps?

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
```diff
@@ -82,60 +120,114 @@ Disabling SSL validation may have undesired side effects and security implicatio
...
+;; SERVER: 127.0.0.53#53(127.0.0.53)
+;; WHEN: Fr Jän 14 11:23:12 CET 2022
+;; MSG SIZE  rcvd: 90
+{{< /command >}}
+
+If the DNS resolves the subdomain to your localhost (127.0.0.1), your setup is working.
+If not, please check the configuration of your router / DNS if the Rebind Protection is active or [enable the LocalStack DNS on your system]({{< ref "dns-server#system-dns-configuration" >}}).
+
+
 ## System DNS configuration
```
simonrw, line 170: I like this section, it's nice to have a proper numbered walkthrough!

simonrw (changes requested):
This is a huge improvement thanks, and most of my comments are suggestions I'll let you decide how to proceed, but the `DNS_LOCAL_NAME_PATTERNS` section _needs_ updating.

content/en/references/configuration.md
```diff
@@ -304,10 +304,11 @@ To learn more about these configuration options, see [DNS Server]({{< ref "dns-s
 
 | Variable | Example Values | Description |
 | - | - | - |
-| `DNS_ADDRESS` | 0.0.0.0 (default) | Address the LocalStack should bind the DNS server on (port 53 tcp/udp). Value `0` to disable.
-| `DNS_SERVER` | 8.8.8.8 (default) | Fallback DNS server for non-modified queries.
-| `DNS_RESOLVE_IP` | 127.0.0.1 | IP address the DNS integration should return as A record for modified queries. This will override any automatic detection of the proper response IP.
-| `DNS_LOCAL_NAME_PATTERNS` | | Names which should be resolved to the LocalStack IP, as python-compatible regex.
+| `DISABLE_TRANSPARENT_ENDPOINT_INJECTION` | `0` (default in Pro) \| `1` | Whether to disable DNS resolution of AWS hostnames to the LocalStack container. Pro feature. (see [Transparent Endpoint Injection]({{< ref "user-guide/tools/transparent-endpoint-injection" >}}))
+| `DNS_ADDRESS` | `0.0.0.0` (default) | Address the LocalStack should bind the DNS server on (port 53 tcp/udp). Value `0` to disable.
+| `DNS_SERVER` | `8.8.8.8` (default) | Fallback DNS server for non-modified queries.
```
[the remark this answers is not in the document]
joe4dev: better 👍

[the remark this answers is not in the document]
joe4dev: I also don't like `8.8.8.8` as default because that's not really true. Only if no default upstream DNS is detected.

content/en/references/configuration.md
```diff
@@ -304,10 +304,11 @@ To learn more about these configuration options, see [DNS Server]({{< ref "dns-s
...
+| `DNS_RESOLVE_IP` | `127.0.0.1` | IP address the DNS integration should return as A record for modified queries. This will override any automatic detection of the proper response IP.
```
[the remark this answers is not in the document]
joe4dev: I adjusted "for queries handled by LocalStack" accordingly

content/en/references/configuration.md
```diff
@@ -304,10 +304,11 @@ To learn more about these configuration options, see [DNS Server]({{< ref "dns-s
...
+| `DNS_LOCAL_NAME_PATTERNS` | `.*(ecr\|lambda).*.amazonaws.com` (example) | Skiplist of hostnames that should *NOT* be resolved to the LocalStack container, as a comma-separated list of Python-flavored regex patterns.
```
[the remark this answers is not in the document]
joe4dev: haha, nice 💡 

Let's stick to one preference :)

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
joe4dev: I don't like "technical limitation", it's a security feature ;)

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
joe4dev: I think this option does indeed refer to `DNS_LOCAL_NAME_PATTERNS` because skipping LocalStack resolution enables connecting to real AWS 🤔 (at least in it's original use)

I agree that it also fits for disabling transparent endpoint injection.

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
joe4dev: @dfangl knows all the legacy history once he is back ;)

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
joe4dev: better 👍 
or even more immediately actionable: "Remember to save the default configuration and restore it after testing.

content/en/references/configuration.md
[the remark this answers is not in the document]
simonrw: Mm good point. Maybe 

```markdown
| `DNS_SERVER` | system or `8.8.8.8` | Fallback DNS server for queries not handled by LocalStack. Uses your system DNS if possible, or `8.8.8.8` otherwise.
```

?

content/en/references/configuration.md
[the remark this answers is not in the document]
simonrw: I think this is a confusing artifact of the default value having meaning, rather than being blank. The actual behaviour is to return `DNS_RESOLVE_IP` if `DNS_RESOLVE_IP != 127.0.0.1`, and use the DNS server's logic to look up the value if not (i.e. its internal mappings or from an upstream). This is a bit strange, since the value returned may not be `127.0.0.1` (e.g. when querying `localhost.localstack.cloud` from a correctly configured docker container or lambda function). Hence why I used "is customised".

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
```diff
@@ -8,52 +8,88 @@ aliases:
...
+LocalStack includes a DNS server that enables seamless connectivity to LocalStack from different environments using `localhost.localstack.cloud` (Community + Pro).
+The DNS server is available on all IPv4 addresses within the LocalStack container (i.e., listening to `0.0.0.0`) and resolves `localhost.localstack.cloud` to the LocalStack container.
+Therefore, containers that are configured to use the DNS server can reach LocalStack using `localhost.localstack.cloud`.
+This configuration happens automatically for containers created by LocalStack, including compute resources such as Lambda, ECS, and EC2.
+Your container can be configured to use the DNS server as demonstrated in the [Network Troubleshooting guide]({{< ref "references/network-troubleshooting/endpoint-url#from-your-container" >}}).
+If you wish to use the DNS server on your host system, follow the instructions under [System DNS configuration]({{< ref "dns-server#system-dns-configuration" >}}).
+
+LocalStack Pro additionally offers a transparent execution mode (active by default), which enables seamless connectivity to LocalStack without changing your application code.
```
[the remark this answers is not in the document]
joe4dev: fully agree, pro standardize 👍 

I think we don't have any standard and "transparent execution mode" vs "Transparent Endpoint Injection" also triggered me somewhat. With the separation of "Transparent Endpoint Injection" (noun-feature) and "DNS server" (separate feature), this makes a lot of sense 💯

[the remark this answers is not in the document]
joe4dev: 👉  going forward with "Transparent Endpoint Injection"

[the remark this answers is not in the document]
joe4dev: at least it clarifies better if someone does not understand the baseline 👍

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
```diff
@@ -8,52 +8,88 @@ aliases:
...
 DNS_ADDRESS=0
 ```
 
 {{< alert title="Warning" color="warning" >}}
-We do not recommend this configuration since this disables resolving `localhost.localstack.cloud` to the LocalStack container.
+We do not recommend disabling the DNS server since this disables resolving `localhost.localstack.cloud` to the LocalStack container.
 {{< / alert >}}
 
-You can also specify which exact URLs should be redirected to LocalStack by defining a hostname regex like:
+This option is primarily used by [LocalStack developers]({{< ref "contributing/development-environment-setup" >}}) in host mode because binding port 53 requires root privileges and port 53 might be occupied.
```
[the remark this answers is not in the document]
joe4dev: moved to the end to de-emphasize following the order discussion above

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
joe4dev: omg 🤦 yess 💯 
I looked up the implementation and messed it up getting confused every time I try to read the config name 🙃 

Fixed 👍

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
joe4dev: fixed `localhost` and re-phrased this part in the context of TEI

content/en/references/configuration.md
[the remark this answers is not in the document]
joe4dev: very good point. Thanks 👍 
(this dynamic behavior is still not covered well and people might be surprised why it works from within a Lambda container)

sadly badly confusing 😕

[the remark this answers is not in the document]
joe4dev: re-added the "customized" part (us-en) ;)

content/en/references/configuration.md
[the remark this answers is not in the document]
joe4dev: I went for "Default upstream DNS or `8.8.8.8` (default)"

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
```diff
@@ -8,52 +8,88 @@ aliases:
...
+
+If you do not want Lambda functions to use the transparent endpoint execution mode in LocalStack Pro, opt out using:
+
+```bash
+DISABLE_TRANSPARENT_ENDPOINT_INJECTION=1
+```
+
+This option disables DNS resolution of AWS domains to the LocalStack container and prevents Lambda from disabling SSL validation.
+With disabled transparent endpoint execution mode, the AWS SDK within Lambda functions might connect to the real AWS API.
+The transparent endpoint execution mode is only available in LocalStack Pro.
```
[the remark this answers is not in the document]
joe4dev: added this pointer to the TEI page 👍

content/en/user-guide/tools/transparent-endpoint-injection/_index.md
```diff
@@ -3,26 +3,88 @@ title: "Transparent Endpoint Injection"
...
+For example, the [`boto3` AWS SDK for Python](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.client) provides a parameter `verify=False` to disable SSL verification.
+Similar parameters are available for most other [AWS SDKs](https://docs.aws.amazon.com/sdkref/latest/guide/version-support-matrix.html).
+
+For Node.js, you can set this environment variable in your application, to allow the AWS SDK to talk to the local APIs via SSL:
+
+```node.js
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"
+```
+
+If you are using the Java AWS SDK v2 in Lambda, you can opt in to automatically disable SSL validation using the configuration `LAMBDA_DISABLE_JAVA_SDK_V2_CERTIFICATE_VALIDATION=1`.
```
joe4dev, line 85: @dfangl @dominikschubert sadly, we missed making this breaking change upon v3 announced at the 2.3.0 release:
> Transparent endpoint injection for Lambdas running AWS Java SDK v2 is now available with the feature flag LAMBDA_DISABLE_JAVA_SDK_V2_CERTIFICATE_VALIDATION=1 . This will be default behaviour by 3.0.
https://discuss.localstack.cloud/t/localstack-release-v2-3-0/533#other-changes-and-enhancements-14

Should we add it to the Lambda configurations (so we won't miss it at v4)?

joe4dev: *general config discussion*: do we strive to have all defaults `0` for configurations?

If we ever change the default for a single config, it gets inconsistent. Hence avoiding double negations might be clearer in most cases 💭

content/en/user-guide/tools/transparent-endpoint-injection/dns-server.md
joe4dev: FYI: I ended up re-phrasing it in the new [LocalStack endpoints section](https://docs.localstack.cloud/user-guide/tools/dns-server/#localstack-endpoints) based on @simonrw suggestion of mentioning `LOCALSTACK_HOST`.

@dfangl feel free to chime in if you remember what the real purpose was 🪄 :)

PR #945, opened 2023-12-01
Update installation.md

anentropic:
Fix typo(?) where docker-compose config should fall back to default value if env var not set, I think

### DIFF
--- a/content/en/getting-started/installation.md
+++ b/content/en/getting-started/installation.md
@@ -244,7 +244,7 @@ version: "3.8"
 
 services:
   localstack:
-    container_name: "${LOCALSTACK_DOCKER_NAME-localstack-main}"
+    container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
     image: localstack/localstack
     ports:
       - "127.0.0.1:4566:4566"            # LocalStack Gateway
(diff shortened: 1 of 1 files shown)


joe4dev:
Thank you for the suggestion @anentropic 

Were you able to notice any issues in your scenario?

According to the Docker documentation, both `-` and `:-` are valid interpolations with default, just with slightly different semantics regarding empty values: https://docs.docker.com/compose/environment-variables/env-file/#interpolation

> Default value
>
>    ${VAR:-default} -> value of VAR if set and non-empty, otherwise default
>    ${VAR-default} -> value of VAR if set, otherwise default

Hence, `-` will pass empty values, and `:-` won't.

anentropic:
I think I ended up with a container named as `-localstack-main`

I'd interpreted that as resulting from concatenating empty `$LOCALSTACK_DOCKER_NAME` with `-localstack-main` when the default value was supposed to be `localstack-main` without the leading hyphen

but I can see that's not what the code does either, and I wasn't aware of the second interpolation syntax

so I'm not sure what's up

joe4dev:
> I think I ended up with a container named as -localstack-main

mhm, I didn't expect that and cannot reproduce it with `Docker Compose version v2.23.0-desktop.1`

`LOCALSTACK_DOCKER_NAME="" docker compose up` results in `localstack-1`

anentropic:
hmm, I don't reproduce it either now... I get `localstack-main` as intended

I guess maybe I was having a bad day and problem was on my side somewhere 😅

PR #961, opened 2023-12-13
transcribe: add note for support of only audio file transcription

sannya-singal:


viren-nadkarni (changes requested):
Could you also fix the MKV file format (remove 'video' from it)?

ackdav:
if we support mp4, why do we only support audio?

sannya-singal:
> if we support mp4, why do we only support audio?

The initial version was focused to work on audio files only. We need to handle multiple streams for video files. Maybe @viren-nadkarni can give more inputs here who previously worked on developing this.

viren-nadkarni (approved):

PR #971, opened 2023-12-18
add note about ec2 termination

HarshCasper:


content/en/user-guide/aws/elastic-compute-cloud/index.md
```diff
@@ -172,6 +172,11 @@ $ ssh -p 12862 -i key.pem root@127.0.0.1
 {{< alert title="Hint" color="success">}}
 If the `ssh` command throws an error like "Identity file not accessible" or "bad permissions", then please make sure that the key file has a restrictive `0400` permission as illustrated [here](#create-a-key-pair).
 {{< /alert >}}
+
+{{< alert title="Note">}}
+LocalStack does not clean up any Docker containers it has created. To terminate the EC2 instance, you will need to make a [`TerminateInstances`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_TerminateInstances.html) API call to stop the Docker container before stopping the LocalStack instance.
```
viren-nadkarni, line 177: I suggest something like follows to make is more clear:

```suggestion
LocalStack does not clean up any Dockerized instance when shutting down.
The Docker containers backing the EC2 instance continue running even after LocalStack is stopped.
You need to explicitly call the [`TerminateInstances`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_TerminateInstances.html) API to stop and remove these containers before stopping the LocalStack.
```

viren-nadkarni (changes requested):

PR #977, opened 2023-12-21
fix api gateway doc

HarshCasper:


bentsku (approved):
LGTM! Thanks for fixing this!

PR #979, opened 2023-12-21
move lambda tools to user guides

HarshCasper:


content/en/user-guide/lambda-tools/_index.md
```diff
@@ -1,11 +1,12 @@
...
-weight: 10
+weight: 45
 description: >
   Develop your Lambdas more efficiently.
 aliases:
   - /tools/lambda-tools/
+  - /user-guide/tools/lambda-tools/
 ---
 
 Lambdas are awesome!
```
joe4dev, line 12: That should be more informative ;)

content/en/user-guide/lambda-tools/lambda-vscode-extension/index.md
```diff
@@ -1,8 +1,10 @@
 ---
 title: "Lambda VSCode Extension"
```
joe4dev, line 2: The "Lambda" prefix is not really needed here

joe4dev: same in the text below

joe4dev, line 2: IMHO, the "LocalStack VSCode Extension" would fit better under "LocalStack Tools". In its current state, it is indeed very limited and only relevant for Lambda, I guess that's why it ended up under the "Lambda Tools":

Screenshot 2024-01-02 at 19 14 06

joe4dev (approved):
I added a few comments/suggestions but it looks fine overall 👍 
It helps that the "Lambda Tools" are right below the "LocalStack Tools" such that looking for them under "LocalStack Tools" likely leads to finding them anyway. Even if alphabetically ordered.

content/en/user-guide/lambda-tools/lambda-vscode-extension/index.md
HarshCasper: I believe we can still keep it around in `Lambda Tools` given that it only focusses on improving the Lambda Development experience right now.

dfangl (approved):
I do not have a definite opinion about the moving of the pages one level up, but it is fine by me!

PR #981, opened 2023-12-21
deprecate cockpit docs

HarshCasper:


Pive01 (commented):
Thanks for tackling this :muscle: !
I feel like the Alert should be on top of the page somewhere close to the tile as it should be the first thing you see and not "hidden" under an image. And as a second comment, since we are in the desktop page I would rephrase it with something like "LocalStack Desktop now replaces Cockpit. ...."
image

PR #982, opened 2023-12-22
add docs for s3 docker image

HarshCasper:


content/en/user-guide/aws/s3/index.md
```diff
@@ -211,6 +211,45 @@ $ awslocal s3api get-bucket-cors --bucket cors-bucket
...
+      - "127.0.0.1:4566:4566"            # LocalStack Gateway
+    environment:
+      - DEBUG=${DEBUG:-0}
+    volumes:
+      - "${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack"
+      - "/var/run/docker.sock:/var/run/docker.sock"
+{{< /tab >}}
+{{< tab header="Docker" lang="shell" >}}
+docker run \
+  --rm -it \
```
bentsku, line 245:
```suggestion
  --rm \
```
I don't think we need the `interactive` mode for running LocalStack, as we don't pass any command to it.

bentsku (approved):
LGTM! Just a minor comment regarding the `docker run` command, but this is great. 
Thanks a lot for taking care of this, I'm really grateful.

I believe that as the image is already out, it's better for us to have documentation about it, it's easier to link to it if people are asking questions.

PR #983, opened 2023-12-22
  remove extensions from references

HarshCasper:


thrau (approved):
As discussed offline, I'm not 100% convinced this is the right way to go, but respect your opinion to clean up the docs and remove redundancies.

The document in references was deliberately information-oriented, whereas the user guide was deliberately learning-oriented. i like the distinction, even though it may create some duplication. there was always a way to get to the user guides from the references. in the future i would also like to have an API doc there which describes the extension API, without a user guide.

PR #984, opened 2023-12-22
docs for quarkus

HarshCasper:


content/en/user-guide/integrations/quarkus/index.md
```diff
@@ -0,0 +1,156 @@
...
+Both Lambda clients (sync and async) can be configured through the `application.properties` file, which should be located in the `src/main/resources` directory. Additionally, ensure that a suitable implementation of the sync client is added to the `classpath`. By default, the extension employs the URL connection HTTP client, so it's necessary to include a URL connection client dependency in the `pom.xml` file:
+
+```xml
+<dependency>
+    <groupId>software.amazon.awssdk</groupId>
+    <artifactId>url-connection-client</artifactId>
+</dependency>
+```
+
+If you want to use Apache HTTP client instead, configure it as follows:
```
whummer, line 58:
```suggestion
If you want to use Apache HTTP client instead, configure it as follows in `application.properties`:
```

content/en/user-guide/integrations/quarkus/index.md
```diff
@@ -0,0 +1,156 @@
...
+- [SQS](https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-sqs.html)
+- [SNS](https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-sns.html)
+- [SES](https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-ses.html)
+- [Secrets Manager](https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-secretsmanager.html)
+- [KMS](https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-kms.html)
+- [IAM](https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-iam.html)
+
+## Configuration
+
+The Configuration properties are fixed at build time. All the other configuration properties are overridable at runtime.
```
whummer, line 120:
```suggestion
The following configuration properties are fixed at build time. All the other configuration properties can be overridden at runtime.
```

content/en/user-guide/integrations/quarkus/index.md
```diff
@@ -0,0 +1,156 @@
...
+| `quarkus.aws.devservices.localstack.container-properties` | `Map<String,String>`  |                                      |
+| `quarkus.aws.devservices.localstack.additional-services."additional-services".enabled` | boolean                |                                      |
+| `quarkus.aws.devservices.localstack.additional-services."additional-services".shared`  | boolean                | `false`                              |
+| `quarkus.aws.devservices.localstack.additional-services."additional-services".service-name` | string                 | `localstack`                         |
+| `quarkus.aws.devservices.localstack.additional-services."additional-services".container-properties` | `Map<String,String>`  |                                      |
+
+{{< alert title="Note" >}}
+- If `quarkus.aws.devservices.localstack.additional-services."additional-services".enabled` is set to `true` and the endpoint-override is not configured, LocalStack will be started and utilized instead of the provided configuration. For all services excluding Cognito, LocalStack will function as the core cloud emulator.  In the case of Cognito, the emulation/mocking will be done by Moto.
```
whummer, line 135: Not sure whether the additional information related to Moto is required - seems like an implementation detail and not really necessary information for the user, wdyt? 

Also, curious why we're using Cognito from moto - presumably to allow users to run local tests even with the community version..? If that is the case, then we could maybe mention that Moto provides a basic version, whereas the full-fledged features are provided when using LS Pro?

content/en/user-guide/integrations/quarkus/index.md
```diff
@@ -0,0 +1,156 @@
...
+- In `quarkus.aws.devservices.localstack.additional-services."additional-services".service-name`, the value of the `quarkus-dev-service-localstack` label is attached to the initiated container. In dev mode, when the shared flag is true, Dev Services checks for a container with the quarkus-dev-service-localstack label set to the configured value before starting a new one. If found, it utilizes the existing container. Otherwise, it creates a new container with the `quarkus-dev-service-localstack` label set to the specified value. In test mode, Dev Services groups services with the same service-name value into a single container instance. This property is useful when there's a requirement for multiple shared LocalStack instances.
```
whummer, line 137:
```suggestion
- In `quarkus.aws.devservices.localstack.additional-services."additional-services".service-name`, the value of the `quarkus-dev-service-localstack` label is attached to the initiated container. In dev mode, when the shared flag is true, Dev Services checks for a container with the `quarkus-dev-service-localstack` label set to the configured value before starting a new one. If found, it utilizes the existing container. Otherwise, it creates a new container with the `quarkus-dev-service-localstack` label set to the specified value. In test mode, Dev Services groups services with the same service-name value into a single container instance. This property is useful when there's a requirement for multiple shared LocalStack instances.
```

content/en/user-guide/integrations/quarkus/index.md
```diff
@@ -0,0 +1,156 @@
...
+weight: 11
+description: >
+  Use the Quarkus framework with LocalStack
+---
+
+Quarkus is a Java framework optimized for cloud, serverless, and containerized environments. Quarkus leverages a Kubernetes Native Java stack tailored for GraalVM & OpenJDK HotSpot, which further builds on various Java libraries and standards.
+
+Localstack is supported by Quarkus as a Dev service for Amazon Services. Quarkus Amazon Services automatically starts a LocalStack container in development mode and when running tests, and the extension client is configured automatically.
+
+{{< alert title="Note" >}}
```
whummer, line 12: nit: seems like quite a specific detail - this is the first mention of `pom.xml` in the text and at this point the user doesn't even know about the Maven setup yet. Perhaps it could be moved a bit further down the page?

whummer (approved):
Kudos for adding this @HarshCasper ! 🚀 LGTM overall, just a few smaller comments. Would be good to get another pair of eyes on it maybe. 👀  Also, I didn't actually try running the Quarkus integration, but will try to give it a whirl in the upcoming days..

tinyg210:
@whummer it's in progress 👍🏻

tinyg210:
Note: it seems that the Quarkus documentation has not been fully tested or is outdated. I've flipped the lambda tutorial upside down, and somehow, the Dev Services are not jumping into action. Will open a PR with them to let them know there are issues

content/en/user-guide/integrations/quarkus/index.md
HarshCasper: Sure, will do 👍

content/en/user-guide/integrations/quarkus/index.md
HarshCasper: > Also, curious why we're using Cognito from moto

We aren't recommending it 😅 Quarkus uses Moto for Cognito for some reason, and not LocalStack. It maybe because Cognito falls under LocalStack Pro.

alexrashed (changes requested):
Just chiming in here to make sure this isn't merged in the current state, because it seems like this PR is currently not properly rebased on `main` (since it contains lots of commits)...

HarshCasper:
> Just chiming in here to make sure this isn't merged in the current state, because it seems like this PR is currently not properly rebased on `main` (since it contains lots of commits)...

Yes, this has been fixed now.

HarshCasper:
> Note: it seems that the Quarkus documentation has not been fully tested or is outdated. I've flipped the lambda tutorial upside down, and somehow, the Dev Services are not jumping into action. Will open a PR with them to let them know there are issues

@tinyg210 I'm merging this presently as it is, since the bug fixing would require someone from Quarkus to look into it. We can circle back and update the documentation later.

PR #990, opened 2023-12-27
add docs for eksctl

HarshCasper:


dfangl (approved):
I think this is a nice start of the docs!
Not sure if we might want to mention it, but there are still some operations not yet supported of eksctl, only the basic cluster creation works.

PR #996, opened 2024-01-01
remove weight from integrations

HarshCasper:


MarcelStranak:
Thanks. It's looking better now.

Can we put the `Language SDKs` menu at the top or bottom of the menu? At present, it is situated at the bottom of the list. I would prefer it to be at the top.

image

lakkeger:
@MarcelStranak it's a good idea to group them together, however I'd simply keep CDK for Terraform just like Hashicorp does, as a subsection/subcategory of Terraform.

HarshCasper:
> @MarcelStranak it's a good idea to group them together, however I'd simply keep CDK for Terraform just like Hashicorp does, as a subsection/subcategory of Terraform.

In this case, I will move CDK for Terraform into the Terraform docs 👍

Issue #224, opened 2022-08-05
add docs for lambda event source mappings


thrau:
add a section on event source mappings to https://docs.localstack.cloud/aws/lambda/

* which event sources does localstack support?
* what are their limitations?
* do they have any special configuration?

PR #1002, opened 2024-01-04
document supported ESMs for Lambda

HarshCasper:
Fixes https://github.com/localstack/docs/issues/224

### DIFF
--- a/content/en/user-guide/aws/lambda/index.md
+++ b/content/en/user-guide/aws/lambda/index.md
@@ -103,13 +103,14 @@ The following output would be retrieved:
 The product of 10 and 10 is 100% 
 ```
 
-## Special Tools
+## Lambda Event Source Mappings
 
-LocalStack provides various tools to help you develop, debug, and test your AWS Lambda functions more efficiently.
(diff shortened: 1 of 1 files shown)


content/en/user-guide/aws/lambda/index.md
```diff
@@ -232,6 +232,15 @@ However, the new provider still supports the following configuration options:
 * The `LAMBDA_DOCKER_DNS` option since LocalStack 2.2.
 * The `HOSTNAME_FROM_LAMBDA` option since LocalStack 3.0.
 
+## Lambda Event Source Mapping
```
joe4dev, line 235: Mappings

joe4dev, line 235: Consider changing the order:
Screenshot 2024-01-04 at 19 15 14

a) Event Source Mapping is an AWS feature and should probably after "Getting Started" rather than at the bottom
b) "Special Tools" might fit better before the RB so we cover the AWS features first and then provide the LocalStack-specific added value like special tools and RB.

joe4dev (approved):
LGTM, added some suggestions regarding the order

PR #1009, opened 2024-01-08
minor fix to lambda hot-reloading docs

giograno:
While I was looking into restoring a state with a hot-reloading lambda function, I noticed that the function we mention is outdated/misplaced (see https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/python/example_code/lambda/lambda_handler_basic.py).

This small fix would make the sample work.


joe4dev (approved):
LGTM 👍

Issue #1022, opened 2024-01-18
Quickstart page is missing port in the URL to launch the webapp


tnarik:
Tiny issue, the `Run the sample AWS application` section is missing the *port* (4566), so it doesn't work out of the box.

It is not hard to figure out, and the localstack-workshop project contains the correct URL, but it would be helpful for newcomers if the port is included in the quickstart tutorial page

tnarik:
I see it has been already updated. thanks so much.

PR #1020, opened 2024-01-16
Docs' improvements

bhavya3024:
1. Add info for in order to run hugo, go is also required
2. Need for append port in url otherwise it doesn't work. (more info in PR comments).
@lizconlan @simonrw @fictorial @SimonWallner

PR #1023, opened 2024-01-18
add url port in quickstart sample

ackdav:
this PR adds the port to the URL within the quickstart sample

if ok, feel free to directly merge

resolves: #1022


simonrw:
Thanks for your update. This change is included in #1020 so I will close this in favour of that one.

PR #1032, opened 2024-01-24
fix stack re-creation tip

alexrashed:
Slightly rephrases the sentence on preferring stack recreations (instead of updating).


alexrashed:
Closed because it's a duplicate.

PR #1033, opened 2024-01-24
re-enable edit link

alexrashed:
## Motivation
In the past, we decided to remove the edit button from the docs pages.
But recently, we have seen that it might be useful to have a fast and easy way to propose changes to the docs.
This PR re-enables the edit button, which has been disabled in a rework of the sidebar in https://github.com/localstack/docs/pull/528 by @lukqw.
This should be used as the start of a discussion, and we can use the preview environment to play around with it.
I used the preview environment to create a short screencast of the feature used to create https://github.com/localstack/docs/pull/1031:
[Screencast.webm](https://github.com/localstack/docs/assets/2796604/a5ab8287-2588-4787-83ed-c8e054f3b040)

~~I explicitly made this a draft PR for now to make sure that we explicitly decide on re-enabling this feature.~~
Since there isn't really an argument against re-enabling the button, I'm changing the PR from draft to ready fro review.

## Changes
- Re-enables the "Edit this page" link on each page.

### DIFF
--- a/layouts/partials/page-meta-links.html
+++ b/layouts/partials/page-meta-links.html
@@ -43,9 +43,9 @@
     
     
   <div class="toc-title">Improve our documentation</div>
-  {{/*  <a href="{{ $viewURL }}" class="td-page-meta--view" target="_blank" rel="noopener"><i class="fa-solid fa-file-lines fa-fw"></i> {{ T "post_view_this" }}</a>
+  {{/*  <a href="{{ $viewURL }}" class="td-page-meta--view" target="_blank" rel="noopener"><i class="fa-solid fa-file-lines fa-fw"></i> {{ T "post_view_this" }}</a>  */}}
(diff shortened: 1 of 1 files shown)


lukqw (commented):
Thanks a lot for being so pro-active about allowing easier editing of our docs!

I no longer remember the concrete reason, but I think back when this was removed it was mainly to reduce the amount of clutter in the side bar.

Given the ease-of-use editing this provides, I think it would still be a net improvement overall.

As long as committing directly to the main branch is disabled, we should be good 👍 
Screenshot 2024-01-25 at 13 01 03

alexrashed:
@lukqw Good point! 👍🏽 
This was just possible due to missing branch restrictions, which were overdue anyways. I added them, and now you cannot commit directly to main anymore.

HarshCasper (approved):

PR #1036, opened 2024-01-24
rephrase wording in section about self assigning licenses

lukqw:
Rephrasing parts of our docs around self assigning licenses to make them more clear


SimonWallner (approved):
:rocket:

PR #1042, opened 2024-01-26
remove v3 notice

HarshCasper:


silv-io (approved):
LGTM!

alexrashed (approved):
Thanks for removing the notice! 🚀💯
Actually, I think we could delete it (instead of encapsulating it in an HTML comment), because we can always look at the commit history (but this is just a nitpick and a subjective opinion).

PR #1047, opened 2024-01-31
Update index.md

davidmadi:


HarshCasper (commented):
Hi,

Any apparent reason to explicitly add the region? `awslocal` defaults to `us-east-1` unless mentioned explicitly.

PR #1058, opened 2024-02-06
Specify Docker Desktop in development-environment-setup.md

bblommers:
The list of required tools mentions `docker`, but that is a umbrella-term for a lot of things. Renaming it to `Docker Desktop` makes it more obvious that this specific Docker tool is required.

The additional Tip is to help people who only have the Docker engine installed.  One might think that the Docker Engine is enough, but newer versions of the engine do not include `buildx` by default, and building the LS Docker image fails in an odd way of `buildx` cannot be found.


HarshCasper (commented):
Hi Bert — Apologies for not looking into this earlier! We have migrated the development environment setup guide to the [LocalStack repo](https://github.com/localstack/localstack). Would you mind to raise a PR over here — https://github.com/localstack/localstack/tree/master/docs/development-environment-setup

PR #1059, opened 2024-02-06
add a changelog page

HarshCasper:
I was looking into docs analytics today and saw some folks have been looking for a Changelog on our docs. Currently, we have changelogs on Discuss for our minor releases. 

This PR:

- Adds a new changelog page
- Links it over the references section
- Documents every single major & minor release since 1.0

### DIFF
--- a/content/en/getting-started/changelog.md
+++ b/content/en/getting-started/changelog.md
@@ -1,0 +1,29 @@
+---
+title: "Changelog"
+linkTitle: "Changelog"
+weight: 70
+description: >
+  This page lists new features, highlights, and bug fixes for official LocalStack releases.
+cascade:
+  type: docs
+---
+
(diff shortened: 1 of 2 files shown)


thrau (commented):
Thanks for tackling this, i think this can be super useful!

Just a few comments:
* PR descriptions would be great to make sure everyone who wasn't part of the discussion and just sees the PR understands what we are doing and why we are doing it :-)
* I'm not sure i really like this in the Getting Started section. What's the motivation for putting it here rather than elsewhere?
* Can we write a template and a data file to generate this list more easily? Would make maintenance easier
* 3.1 seems to link to a particular paragraph, maybe `https://discuss.localstack.cloud/t/localstack-release-v3-1-0/713` would be better?

PR #1064, opened 2024-02-08
add a note about public ecr image

HarshCasper:


MarcelStranak (changes requested):
Please add a subsection title so that it will be visible as another option. Otherwise ok.

PR #1066, opened 2024-02-08
Document ECS_DISABLE_AWS_ENDPOINT_URL

simonrw:
An upcoming feature of ECS is that we will inject `AWS_ENDPOINT_URL` into the users' container environment. This is a change to the defaults, so we must provide a configuration flag (`ECS_DISABLE_AWS_ENDPOINT_URL`) to restore the previous behaviour. 

This change documents that new flag, taking inspiration (i.e. copied from) the `LAMBDA_DISABLE_AWS_ENDPOINT_URL` variable.


joe4dev (approved):
LGTM 👍

PR #1068, opened 2024-02-08
add a note about s3 force path style

HarshCasper:


content/en/user-guide/aws/s3/index.md
```diff
@@ -132,6 +132,10 @@ However, if the endpoint is not prefixed by `s3.`, LocalStack will not be able t
 You can either change the endpoint to an S3-specific one, or configure your SDK to use **Path style** requests instead.
 Check out our [SDK documentation](https://docs.localstack.cloud/user-guide/integrations/sdks/) to learn how you can configure language SDKs to access LocalStack and S3.
 
+{{< alert title="S3 Force Path Style" >}}
+While using AWS language SDKs, you would need to configure `ForcePathStyle` to `true` in the S3 client configuration to use **Path style** requests. The S3 service endpoint is slightly different from the other service endpoints, because AWS is deprecating path-style based access for hosting buckets. If you want to use virtual host addressing of buckets, you can remove `ForcePathStyle` from the configuration. 
+{{< /alert >}}
+
```
bentsku, line 138: Thanks for the PR! I'm not sure we should talk about deprecation because people have been worried about this before: AWS came back on the deprecation and I don't think they ever will be able to remove path style, so many legacy things are based on it for now. I guess we should update this messages from the SDK docs as well.
Also, the ForcePathStyle name is different from SDK to SDK, which can be confusing to talk about it here.

```suggestion
{{< alert title="S3 Force Path Style" >}}
While using AWS language SDKs, you would need to configure the `ForcePathStyle` parameter to `true` in the S3 client configuration to use **Path style** requests. If you want to use virtual host addressing of buckets, you can remove `ForcePathStyle` from the configuration.
The `ForcePathStyle` parameter name can vary between SDK and languages, please check our [SDK documentation](https://docs.localstack.cloud/user-guide/integrations/sdks/)
{{< /alert >}}
```

We could maybe add a link to the AWS SDKs too? https://aws.amazon.com/developer/tools/#SDKs

bentsku (commented):
Thanks a lot for the PR! Maybe we could remove the deprecation part, because AWS has pushed back the date and I'm not sure they will ever really be able to actually deprecate it, and it tends to worry people who then absolutely do not want to use path style.

Added a comment about it, what do you think?

Issue #998, opened 2024-01-02
AWS CDK UPDATE not fully supported


k-chainalysis:
### Actual Behavior

CDK `UPDATE` is not fully supported, but that is not called out in the docs.

Inability to re-deploy after making code changes is a significant difference with expected AWS CDK behavior, so it should be communicated.

### Expected Behavior

If CDK `UPDATE` is not fully supported, the **Limitations** section of *Integrations > AWS CDK* page should call that out.

k-chainalysis:
![2024-01-02_15-18-57](https://github.com/localstack/docs/assets/103948345/19903045-7324-4fbc-a38f-3458b457aed3)

PR #1077, opened 2024-02-12
add a note about updating cdk stacks

HarshCasper:
Fixes https://github.com/localstack/docs/issues/998


k-chainalysis:
Looking forward to expanded `UPDATE` functionality in CDK. Thanks for clarifying docs!

PR #1078, opened 2024-02-12
add FAQ on Docker image tags

alexrashed:
### Motivation
There is some misconception on the semantics of our Docker tags (e.g. "what means `latest`"?).
This PR adds a section to the FAQ section on the docs describing the semantics of the different tags.
Happy for any feedback!

### Changes
- First commit contains some cleanup to have every sentence in the FAQ section in its own line in the markdown file (according to our best practices, which helps a lot with reviews / diffs).
- The second commit adds a section which explains the semantics of the different tags.


thrau (approved):
LGTM! thanks for adding this so quickly :-) this will be super helpful for communication.

in the future we could maybe reconcile this (or simply add it) to the dockerhub docs https://github.com/localstack/localstack/issues/10200 (there's also a community PR already open for this :-))

HarshCasper (approved):

PR #1080, opened 2024-02-14
link official extensions to the web app

HarshCasper:
This PR removes the official extension doc page since it requires manual effort to update and maintain, and instead link it directly to the web app


content/en/user-guide/extensions/official-extensions/_index.html
```diff
@@ -0,0 +1,8 @@
+---
+linktitle: Official Extensions
+title: Official Extensions
+layout: redirect
+sitemap_exclude: true
+weight: 4
+redirect: https://app.localstack.cloud/extensions/library
+---
```
thrau, line 8: nit: can we add a description here? it will show up in the overview page

thrau (approved):
LGTM!

PR #1096, opened 2024-02-26
improve help & support docs

HarshCasper:


komarkovich:
Looks good however we would change the **pro** --> **dedicated** (dedicated support instead of pro support)
I would change the wording here a little bit as well:

> LocalStack Support Team provides comprehensive support for all the Pro users, including the Team and Enterprise plans.

to:

> Dedicated customer support is available to all the LocalStack users with an active subscription.

PR #1159, opened 2024-04-02
switch term 'version' with 'image'

steffyP:
For the coverage api we still need to adapt the terminology, as we currently use the wording `Edition` and `version` which both should be replaced with `image`. 

* This PR updates the templates we have that are used to create the individual pages
* The changes can be seen in the preview: https://localstack-docs-preview-pr-1159.surge.sh/references/coverage/


SimonWallner (approved):
:rocket:

PR #1181, opened 2024-04-17
Update ECS quickstart to demonstrate CloudWatch Logs integration

simonrw:
# Motivation

We recently integrated live log streaming from ECS containers to CloudWatch, rather than dumping the logs at the end of the task (which may never happen normally for services). We should highlight this feature as it improves parity, as well as developer experience.

# Changes

- Add log configuration to the example task definition
- Highlight that logs should be available
- Show example of fetching container logs


pinzon (approved):
Great!

PR #1186, opened 2024-04-22
Specify that S3 signature validation needs the accesskey/secretkey to be set to `test`

krishanbhasin-px:
I've been setting up my localstack S3 and couldn't get the signature validation passing.

I came across [this comment](https://github.com/localstack/localstack/issues/9538#issuecomment-1792255886) which indicated I needed to set my access key/secret key to the specific value of `test` in order to make this work; this PR adds that note to the configuration information for S3.

I had been trying to use a structured S3 key, like referenced [in the docs](https://docs.localstack.cloud/references/credentials/#structured-access-key-id)

I had thought this might be related to me running the `s3-latest` docker image which presumably doesn't include STS, but swapping to the full image doesn't seem to fix it, even when specifying `SERVICES=s3,sts`

### DIFF
--- a/content/en/references/configuration.md
+++ b/content/en/references/configuration.md
@@ -247,7 +247,7 @@ Please consult the [migration guide]({{< ref "user-guide/aws/lambda#migrating-to
 | Variable | Example Values | Description |
 | - | - | - |
 | `S3_DIR` || **Deprecated since 3.0.0** This is only supported for the `legacy_v2` provider. Configure a global parent directory that contains all buckets as sub-directories (`S3_DIR=/path/to/root`) or an individual directory that will get mounted as special bucket names (`S3_DIR=/path/to/root/bucket1:bucket1`). Only available for Localstack Pro.
-| `S3_SKIP_SIGNATURE_VALIDATION`| `0` \| `1` (default) | Used to toggle validation of S3 pre-signed URL request signature. Set to `0` to validate. Note that validation only passes if the AWS_SECRET_KEY and AWS_ACCESS_KEY variables are set to `test` |
+| `S3_SKIP_SIGNATURE_VALIDATION`| `0` \| `1` (default) | Used to toggle validation of S3 pre-signed URL request signature. Set to `0` to validate. Note that validation can only pass if the AWS_SECRET_KEY and AWS_ACCESS_KEY variables are set to `test` |
 | `S3_SKIP_KMS_KEY_VALIDATION` | `0` \| `1` (default) | Used to toggle validation of provided KMS key in S3 operations. |
 | `PROVIDER_OVERRIDE_S3` | `legacy_v2` \| `v3` (default) | The new LocalStack-native S3 provider (v3) is active by default since LocalStack 3.0. |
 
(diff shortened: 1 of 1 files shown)


content/en/references/configuration.md
```diff
@@ -247,7 +247,7 @@ Please consult the [migration guide]({{< ref "user-guide/aws/lambda#migrating-to
 | Variable | Example Values | Description |
 | - | - | - |
 | `S3_DIR` || **Deprecated since 3.0.0** This is only supported for the `legacy_v2` provider. Configure a global parent directory that contains all buckets as sub-directories (`S3_DIR=/path/to/root`) or an individual directory that will get mounted as special bucket names (`S3_DIR=/path/to/root/bucket1:bucket1`). Only available for Localstack Pro.
-| `S3_SKIP_SIGNATURE_VALIDATION`| `0` \| `1` (default) | Used to toggle validation of S3 pre-signed URL request signature. Set to `0` to validate. |
+| `S3_SKIP_SIGNATURE_VALIDATION`| `0` \| `1` (default) | Used to toggle validation of S3 pre-signed URL request signature. Set to `0` to validate. Note that validation can only pass if the AWS_SECRET_KEY and AWS_ACCESS_KEY variables are set to `test` |
```
bentsku, line 250:
```suggestion
| `S3_SKIP_SIGNATURE_VALIDATION`| `0` \| `1` (default) | Used to toggle validation of S3 pre-signed URL request signature. Set to `0` to validate. Note that validation can only pass if the `AWS_SECRET_ACCESS_KEY` is set to `test` or if using credentials returned from `STS.AssumeRole`  |
```

bentsku (changes requested):
Hello @krishanbhasin-px and thanks for your contribution!

I can add a bit more context to it:

If using the regular LocalStack image, you can use the pre-signed URL validation with a created role and getting the credentials for the role, and this will work. 

However, for simplicity purpose, LocalStack treats unrecognized AWS Access Keys as the default `root` account (`000000000000`), and the Secret Access Key is only used for signing purpose.

However, when sending a request to LocalStack with random credentials, LocalStack cannot know what your secret key is. In order to be able to validate the signature, we need to make sure that the secret key used in the client and the "server" (here LocalStack) are the same. This is why the `test` value is hardcoded as a fallback. You can read this PR description to understand a bit more what is happening under the hood: https://github.com/localstack/localstack/pull/10265

This was just to give more context 😄 thanks for adding this, we will also add a note in the S3 page directly to explain a bit more why the Secret Key need to be set. With the linked PR, you now only have to set the Secret Access Key to `test` in order for the validation to be successful, the Access Key Id used will be the one of the request. 

Thanks again for the contribution and sorry for the initial issue!

krishanbhasin-px:
Hey @bentsku that makes much more sense to me now, thank you for taking the time to write such a detailed response.
And thank you for localstack, it's so useful!

bentsku (approved):
@krishanbhasin-px thank you again for the contribution and raising this issue, I'm sorry again it wasn't very clear. 

Also, sorry in the delay in reviewing 😕
Welcome to the contributors! 

I'll open a PR soon trying to condense the previous response into the S3 documentation as well.

rpgdev:
@krishanbhasin-px why doesn't the configuration page also makes it explicit that this setting affects the timeout for signed urls? Makes it impossible to find out from just ctrl+f the configuration page.

PR #1209, opened 2024-04-29
add IAM coverage map

HarshCasper:
This PR documents:

- Add supported services
- Add inter service enforcement 
- Add supported policy types and features


pinzon (approved):

dfangl (approved):
Thanks for migrating this! It is not up to date, (some more operations are tested in the meantime), but a good first iteration, let's reiterate over it in the coming weeks, I will try to regenerate those tables!

Issue #1230, opened 2024-04-30
Glue: no definition of job.py


ksanjeev9211:
the tutorial talks about job.py, however, there is no content for job.py.

**awslocal s3 cp job.py s3://glue-test/job.py**

when I run the above command, i get the following error

**The user-provided path job.py does not exist.**

PR #1233, opened 2024-05-02
add a simple glue script to the docs

HarshCasper:
Fix https://github.com/localstack/docs/issues/1230


alexrashed (approved):
Thanks for jumping on this! The script makes it easier to try out the service! 💯

PR #1245, opened 2024-05-06
use CIRCLE_CI_TOKEN to download artifacts

steffyP:
Today the workflow failed as the download of artifacts from CircleCI wasn't possible.
Testing this locally revealed an unauthorized-message, e.g. 

```
curl -X 'GET' "https://circleci.com/api/v2/project/github/localstack/localstack/203915/artifacts" -H 'accept: application/json'
{"message":"You must log in first"}%
```

Leading to the conclusion that the API may require the `Circle-Token` now to be set in the header. 

However, after finishing the fix, it seems the issue was resolved in CircleCI as the [workflow is running again](https://github.com/localstack/docs/actions/runs/8965000681/job/24633829907), without any changes.

# TODO
- [ ] [workflow](https://github.com/localstack/docs/actions/runs/8970476265/job/24634101178) is still failing in the CI, but it worked locally


alexrashed (approved):
Thanks for jumping on this! 💯 It's pretty weird that the API is jumping around with enforcing auth or not...

PR #1253, opened 2024-05-08
add docs for bitbucket

HarshCasper:


lakkeger (approved):
Great article 🍻 Thanks for putting it together.
I'd recommend though to emphasize that Bitbucket support is really experimental.
Additionally might worth mentioning limitations in a known issues/limitations section ie mounting volumes is not supported by Bitbucket hence we provide limited functionality.

PR #1259, opened 2024-05-13
Public documentation about persistence

giograno:
This PR adds automatically generated documentation about persistence support to our docs.

Everything is generated from the Persistence Catalog on Notion.
A [Python script] builds a [JSON file](https://github.com/localstack/docs/blob/persistence-data/data/persistence/coverage.json) in the `data` folder basically collecting the available data on Notion.

This JSON file is used to build an overview table that summarizes the state of persistence support for every service ([here](https://localstack-docs-preview-pr-1259.surge.sh/user-guide/state-management/support/)).

image

Furthermore, we use Hugo taxonomies to add a persistence tag in every service page (e.g., [here](https://localstack-docs-preview-pr-1259.surge.sh/user-guide/aws/cognito/)).

image

A click on the tag redirects to the overview of the persistence support.
To implement these tags, we automatically parse all the service pages and add a new entry in the YAML front matter.
The tag is only displayed for the supported services.

Similarly to the API coverage updates, we have a pipeline that periodically checks for changes (things should change in the Notion table).


HarshCasper (approved):

layouts/partials/taxonomy_terms_article.html
alexrashed, line 1: Actually this partial overwrite should not be necessary, since there is quite advanced native support for taxonomies.
According to the lookup-order, you can just define a taxonomy landing page directly: https://gohugo.io/templates/lookup-order/
You can do so by:
- Removing this file.
- Creating a new file: `layouts/partials/taxonomy/persistence_support.html`
- This file can then just contain an HTML redirect like this:
  ```html
  <!doctype html>
  <html lang="en">
    <head>
        <meta charset="utf-8">
        <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
        <title>Persistence Overview Redirect</title>
        <link rel="canonical" href='{{ "/user-guide/state-management/support/" | relURL }}'>
        <meta name="robots" content="noindex">
        <meta http-equiv="refresh" content='0; url={{ "/user-guide/state-management/support/" | relURL }}'>
    </head>
  </html>
  ```

Your partial here has some additional logic here, but that could directly be done in the script setting the front-matter (not set the taxonomy if you don't want to display it).

alexrashed (commented):
Wow! This is already looking great! Really an awesome spike to make the persistence support more transparent for our users! 🚀 
I only added a few comments to improve the automation and reduce the amount of overwritten partials (because overwritten partials can be quite a pain for upgrades of Docsy or Hugo).
Let me know what you think, happy to assist with any of the changes!

joe4dev:
Great initiative 🚀 More transparency about coverage is always welcomed.

⚠️ I think it's important to raise awareness so everyone knows which data is synchronized from Notion into public user-facing documentation.
This is currently not obvious in Notion. It mainly applies to the free text field "Limitations" (which I would expect to be something internal when editing in Notion). It's good to have the PR review process as a "gatekeeper" ✔️, but it also needs periodic maintenance reminders.

💡 Small idea: We could add a direct link to these coverage docs overview in the PR preview to facilitate reviewing (e.g., whether the free-text field messes up any formatting). These look predictable (apart from the PR number we probably don't know yet 😬 ): https://localstack-docs-preview-pr-1259.surge.sh/user-guide/state-management/support/

content/en/user-guide/state-management/support/index.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Persistence Support for AWS Services
+linkTitle: Persistence Support
+description: Overview of the persistence support across the implemented AWS services
+hide_readingtime: true
+---
+
+## Persistence Support Overview
```
thrau, line 8: Perhaps we could add an explanation of the table here, what the column mean.

layouts/partials/persistence/persistence_table_row.html
```diff
@@ -0,0 +1,18 @@
+{{ $data := .Site.Data.persistence.coverage }}
+
+{{ range $data }}
+<tr>
+    <!-- Service -->
+    <td>{{ .full_name }}</td>
```
thrau, line 6: would it be possible to add a link to the service? like https://localstack-docs-preview-pr-1259.surge.sh/user-guide/aws/acm/

thrau (commented):
Adding two minor things to the review comments. It already looks great!

definitely +1 on Joel's comments!

thrau:
also added @MarcelStranak to add feedback from a Support perspective

MarcelStranak (commented):
- I would suggest renaming the article to “Persistence Coverage” so we maintain consistency with the existing article titled “LocalStack Coverage.”
- Add weight to the title so it is at the top of the list same as "LocalStack Coverage", plus add it as bold.
- Remove the `limitations` column and mention them directly in the service articles.
- Add terminology for the other columns similar to https://localstack-docs-preview-pr-1259.surge.sh/references/coverage/coverage_acm/#terminology
- Add to the article `Tag: [Pro image]` (same as [persistence article](https://localstack-docs-preview-pr-1259.surge.sh/user-guide/state-management/persistence/))


Otherwise, it looks wonderful.

layouts/partials/persistence/persistence_table_row.html
```diff
@@ -0,0 +1,18 @@
+{{ $data := .Site.Data.persistence.coverage }}
+
+{{ range $data }}
```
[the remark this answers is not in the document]
giograno: ✅

layouts/partials/taxonomy_terms_article.html
giograno: Cool suggestion 💯 I am a Hugo newbie so I did not know about this. I also did not particularly like the idea of overwriting the partial.

content/en/user-guide/state-management/support/index.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Persistence Coverage for AWS Services
+linkTitle: Persistence Coverage
+description: Overview of the persistence coverage across the implemented AWS services
+hide_readingtime: true
+weight: 1
```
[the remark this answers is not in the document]
alexrashed: @MarcelStranak explicitly asked for this, are you okay with removing the weights again since @HarshCasper vetos against it?

scripts/persistence/notion/catalog.py
```diff
@@ -0,0 +1,29 @@
...
+DEFAULT_CATALOG_DATABASE_ID = "3c0f615e7ffc4ae2a034f1ed9c444bd2"
+
+
+class PersistenceServiceItem(Page):
+    name = TitlePlainText("Name")
+    status = Status("Persistence")
+    has_test = Checkbox("Persistence Tests")
+    primary_owner = PeopleProperty("Primary Owner")
+    secondary_owner = PeopleProperty("Secondary Owner(s)")
+    limitations = Text("Limitations")
```
alexrashed, line 24: To address @joe4dev's comment, maybe we could rename this column to something like `Limitations (synced with docs)`?

alexrashed (approved):
Awesome! This now looks really great! Thanks for addressing the comments and reducing the coupling to docsy by moving to the native taxonomy support!
I only added a comment to address @joe4dev's very good point on communicating that certain internal docs are synced to public docs.
I don't really think it's super-easy to update the preview deployment link. You don't know that at the time of the creation of the PR, and when the PR is created you don't know it's a persistence update PR:
https://github.com/localstack/docs/blob/50daf2775a11b9a41ca6a692a9c5c3976792daca/.github/workflows/preview-deploy.yml#L37-L45

So from my point of view: As soon as the weight is adjusted such that @HarshCasper accepts the PR, and the sync is communicated, we're good to go!
Really an awesome new feature of our docs! 🚀

HarshCasper (approved):

PR #1274, opened 2024-05-17
Mount docker socket in installation example

simonrw:
# Motivation

Our getting started example does not mount the Docker socket. If the user does this, they get lots of log messages regarding failed connections to the Docker socket (inside the container). This is off-putting for a new user.

# Changes

* Show mounting the Docker socket into the LocalStack container for the Docker installation section
* Expose ports on 127.0.0.1 by default
* Make more complete examples for Docker and Docker Compose
* Add info about `LOCALSTACK_` prefix for config variables in configuration reference
* Improve wording around why we mount the Docker socket

### DIFF
--- a/content/en/getting-started/installation.md
+++ b/content/en/getting-started/installation.md
@@ -334,6 +334,7 @@ docker run \
   --rm -it \
   -p 4566:4566 \
   -p 4510-4559:4510-4559 \
+  -v /var/run/docker.sock:/var/run/docker.sock
   localstack/localstack
 {{< /tab >}}
 {{< tab header="Pro" lang="shell" >}}
@@ -342,6 +343,7 @@ docker run \
   -p 4566:4566 \
   -p 4510-4559:4510-4559 \
   -e LOCALSTACK_AUTH_TOKEN=${LOCALSTACK_AUTH_TOKEN:?} \
+  -v /var/run/docker.sock:/var/run/docker.sock
   localstack/localstack-pro{{< /tab >}}
 {{< /tabpane >}}
 
--- a/content/en/getting-started/installation.md
+++ b/content/en/getting-started/installation.md
@@ -294,7 +294,7 @@ $ docker-compose up
 
 - This command reuses the image if it's already on your machine, i.e. it will **not** pull the latest image automatically from Docker Hub.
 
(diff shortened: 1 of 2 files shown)


alexrashed (commented):
Thanks for jumping on this! LocalStack really relies on the Docker Socket right now and this should be part of the examples.
We need to make sure that these samples work for basically all our users. Would this command work exactly the same on Windows, Linux, and MacOS?
I would also like to get @HarshCasper's opinion on this :)

HarshCasper:
@alexrashed I'm good with this change. Perhaps, we need a few lines to share why we exactly need to mount the Docker socket, and what are the alternatives if we don't wish to do that.

MarcelStranak:
> Thanks for jumping on this! LocalStack really relies on the Docker Socket right now and this should be part of the examples. We need to make sure that these samples work for basically all our users. Would this command work exactly the same on Windows, Linux, and MacOS? I would also like to get @HarshCasper's opinion on this :)

Works for each OS.

We do mention this in the docs, but only as a note under the [docker compose](https://docs.localstack.cloud/getting-started/installation/#starting-localstack-with-docker-compose).

image


There is also a mention under the [docker cli command](https://docs.localstack.cloud/getting-started/installation/#starting-localstack-with-docker). 

We should potentially change how we provide this type of information in the docs.

image

simonrw:
I have tried to unify and clean up the wording of the documentation here. Please take another look to make sure I've captured your thoughts.

content/en/getting-started/installation.md
```diff
@@ -294,7 +294,7 @@ $ docker-compose up
 
 - This command reuses the image if it's already on your machine, i.e. it will **not** pull the latest image automatically from Docker Hub.
 
-- Mounting the Docker socket `/var/run/docker.sock` as a volume is required for the Lambda service. Check out the [Lambda providers]({{< ref "user-guide/aws/lambda" >}}) documentation for more information.
+- Mounting the Docker socket `/var/run/docker.sock` as a volume is required for some services that use Docker to provide the emulation, such as AWS Lambda.
```
alexrashed, line 297: nit: would it make sense to keep the link to the lambda docs?

alexrashed (approved):
Thanks for jumping on this! Since it was clarified that the changes are interoperable, this is looking great! 🚀

PR #1282, opened 2024-05-22
add notes on CLI version for s3 remote

giograno:
It happened a couple of times that customers could not work with s3 remotes on older CLI versions (mostly the `load` and `list` commands were targeting our platform instead). Since updating the CLI was an easy fix, I added a callout in the docs.


MarcelStranak:
@giograno, just FYI. You can only use that command when the CLI is installed through PIP.
Otherwise, you get the error:

> ❌ Error: The LocalStack CLI can only update itself if installed via PIP. Please follow the instructions on https://docs.localstack.cloud/ to update your CLI.

I would simply direct them to the LocalStack documentation at https://docs.localstack.cloud/getting-started/installation/#localstack-cli, where all the installation options are listed.

HarshCasper (approved):

PR #1295, opened 2024-05-24
add gremlin transaction

cloutierMat:


steffyP (approved):
Great addition to the docs 🚀 
Thanks for adding the sample!

Just FYI for future PRs: we have some guidelines on the formatting and would like to have each sentence in a new line. It makes reading easier, and doesn't have any affect on the resulting output :)

PR #1307, opened 2024-05-30
add missing backslash to Docker settings file to correct path.

C-Williams:
While trying to solve a DNS issue, I found a typo in the docs. Just a simple missing backslash to escape the space.

### DIFF
--- a/content/en/user-guide/tools/dns-server/index.md
+++ b/content/en/user-guide/tools/dns-server/index.md
@@ -155,7 +155,7 @@ sudo lsof -P -i :53 | grep LISTEN
 
 In macOS, a common process that listens on port 53 is `mDNSResponder`.
 Docker for Mac 4.24 has a [known issue](https://docs.docker.com/desktop/release-notes/#4240) and suggests the following workaround:
-> Deactivate network acceleration by adding `"kernelForUDP": false`, in the `settings.json` file located at `~/Library/Group Containers/group.com.docker/settings.json`.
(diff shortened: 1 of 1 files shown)


simonrw:
Hi @C-Williams, thanks for this contribution! 

I recognise that adding a backslash will make it easier for users to copy and paste the path into their terminal/editor, but strictly the space _without_ escaping backslash is correct. This is how paths are represented.

C-Williams:
That's fair. 

I think for me, having the file path written as code (i.e. `like so`), kind of implies that the intention to copy/paste this line. Throughout your docs, you all routinely use `` to represent code, or text seen in the Terminal and not general text and so it could be a little unclear what the intention is here.

If the intention is not to copy/paste this, I might suggest writing this path in double quotes like "Internet Sharing" directly below. 

Just my opinion though, so feel free to close this PR without merging if desired.

simonrw:
You raise an interesting question: I don't really have a standard for when something should be presented in monospace. I usually use monospace for "computer-ey" things like commands, or anything that is represented in the terminal, however I don't personally associate copying and pasting with monospace fonts.

I think the text as represented in our documentation is technically correct, that is the path. The space makes it a problem for copying and pasting, but an alternative solution is to wrap the path in quotes, which would prevent the shell from splitting the path into two parts as well.

If we take inspiration from [Docker's own documentation on this feature](https://docs.docker.com/desktop/settings/mac/), they also don't escape the space, and they _also_ use monospace font to represent the path.

image

Thanks for reaching out and proposing this change, but I will close this PR without merging. I really appreciate you trying to make our documentation better, and I hope you are getting on ok with LocalStack!

PR #1311, opened 2024-06-03
document `--env-file` flag in configuration docs

HarshCasper:
Documents https://github.com/localstack/localstack/pull/10880

### DIFF
--- a/content/en/references/configuration.md
+++ b/content/en/references/configuration.md
@@ -61,7 +61,7 @@ Options to configure how LocalStack interacts with Docker.
 
 | Variable | Example Values | Description |
 | - | - | - |
-| `DOCKER_FLAGS` | | Allows to pass custom flags (e.g., volume mounts) to "docker run" when running LocalStack in Docker. |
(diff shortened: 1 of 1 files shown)


dfangl:
Duplicate of #1287? The `DOCKER_FLAGS` itself (the namings are confusing), are only for the flags for starting the LocalStack container via the CLI, and not for any containers LocalStack itself spawns (like `LAMBDA_DOCKER_FLAGS`).
The `DOCKER_FLAGS` always supported `--env-file`. Maybe we should clarify the difference better here.

HarshCasper:
> Duplicate of #1287? The `DOCKER_FLAGS` itself (the namings are confusing), are only for the flags for starting the LocalStack container via the CLI, and not for any containers LocalStack itself spawns (like `LAMBDA_DOCKER_FLAGS`). The `DOCKER_FLAGS` always supported `--env-file`. Maybe we should clarify the difference better here.

Ah, my bad! Closing this PR

PR #1322, opened 2024-06-11
skip duplicated test details for api coverage

steffyP:
Some API coverage details contained the same test (including the same response code) in the list of tests.
This PR fixes this, by not adding duplicated information to the test-details.


steffyP:
Before this change the test details looked like this:

Screenshot 2024-06-11 at 10 24 56

With the change the test details are much cleaner:

Screenshot 2024-06-11 at 10 24 27

PR #1323, opened 2024-06-11
Update golang s3 example to avoid using deprecated bits

drichelson:
Upon [upgrading the v2 aws sdk dependency](https://github.com/dorklyorg/dorkly/commit/3a212cea2099d504ba73c43c678e4bce668a9d9f) I noticed some deprecation warnings. 

Removes deprecated usages of EndpointResolver in favor of specifying BaseEndpoint. 

This change 'works for me' (my test passes) but could use some additional verification for those with different use cases.
See aws docs: https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/#mutable-endpoint

### DIFF
--- a/content/en/user-guide/integrations/sdks/go/index.md
+++ b/content/en/user-guide/integrations/sdks/go/index.md
@@ -64,22 +64,8 @@ func main() {
   awsEndpoint := "http://localhost:4566"
   awsRegion := "us-east-1"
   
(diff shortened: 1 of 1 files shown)


bentsku (approved):
LGTM! Thanks for the delay on that one, and thanks for updating the docs to follow the best practices from the SDK.

PR #1324, opened 2024-06-12
revamp kinesis analytics docs

HarshCasper:
Preview: https://localstack-docs-preview-pr-1324.surge.sh/user-guide/aws/kinesisanalytics/


steffyP (approved):
Thanks so much for reworking the docs and clarifying some of the previous misleading description 🙏 

IMO this is a good compromise until we can provide fixes for v1 and implementation for v2 that are parity tested 👍

PR #1327, opened 2024-06-13
add a note on --host-dns flag

HarshCasper:


dfangl (approved):
Looks good, thanks for adding it!

PR #1335, opened 2024-06-19
fix EC2 create-a-key-pair anchor links

alexrashed:
## Motivation
@Anze1508 made me aware that the EC2 link from the callout to the "Create a key pair" anchor wrongly leads to "/#create-a-key-pair" instead of "#create-a-key-pair"

## Changes
- Fix the links by directly referencing the anchor on the EC2 page using the `relref` shortcode.

## Next steps
This is just a workaround, it would be great if anchor links would work in the `callout` shortcode introduced by @viren-nadkarni.
Unfortunately this is actually a bug in the `alert` shortcode of docsy (of which the `callout` shortcode is a fork of).


viren-nadkarni (approved):
Thank you for fixing this @alexrashed !

For now, I've documented this workaround in the shortcode src.

PR #1339, opened 2024-06-24
Describe Lambda RIE / init config

joe4dev:
## Motivation

With the recent changes in Lambda RIE log levels (https://github.com/localstack/localstack/pull/11067), we should describe the option of how to configure log levels clearly (e.g., if support needs to ask for more detailed logs).

Preview: https://localstack-docs-preview-pr-1339.surge.sh/user-guide/aws/lambda/#localstack-lambda-runtime-interface-emulator-rie

## Changes

* Add a section to describe the Lambda RIE / Init binary with some config options using a disclaimer regarding API stability (considering it internal API for now)
* Fix Lambda config ordering (alphabetically)

## Discussion

* What's the best name?
  * Runtime Interface Emulator (RIE)
  * LocalStack Runtime Interface Emulator (LRIE)
  * Lambda init binary
  * Lambda runtime init
  * ...
* The section title [here](https://localstack-docs-preview-pr-1339.surge.sh/user-guide/aws/lambda/#localstack-lambda-runtime-interface-emulator-rie) seems a little long 🤔 
  * "LocalStack Lambda Runtime Interface Emulator (RIE)"
  * ...


dominikschubert (approved):
LGTM, also thanks for reordering properly

For me it's fine to keep the (long) title as is. I think using RIE makes sense given how much we're actually building on top of it anyway. It's just a somewhat thin layer above it to glue it together for our requirements

PR #1357, opened 2024-07-03
add extensions env var in a separate section

HarshCasper:


thrau (approved):
Thanks, this look good to me!

As always: a bit of context in the PR description doesn't hurt ;-)

PR #1358, opened 2024-07-04
add a note about the --compose-env-file flag

HarshCasper:


content/en/references/configuration.md
```diff
@@ -498,6 +498,8 @@ $ python -m localstack.cli.main --profile=dev config show
 
 {{< callout >}}
 The `CONFIG_PROFILE` is a CLI feature and cannot be used with a Docker/Docker Compose setup.
-You can look at [alternative means of setting environment variables](https://docs.docker.com/compose/environment-variables/set-environment-variables/) for your Docker Compose setups.
+
 For Docker setups, we recommend passing the environment variables directly to the `docker run` command.
+
+You can also set environment variables in a file and pass it to the LocalStack Docker container by using the `--compose-env-file` flag.
```
dfangl, line 504: I think there is a misunderstanding how the `--compose-env-file` works. It only works in `LAMBDA_DOCKER_FLAGS`, `ECS_DOCKER_FLAGS` and the likes.

It is not a command line parameter for the CLI, nor for the docker CLI, nor will it work for the regular `DOCKER_FLAGS` which are used for starting the LocalStack container itself. It is purely a flag for container localstack starts on its own.

dfangl (changes requested):

content/en/references/configuration.md
```diff
@@ -207,7 +207,7 @@ Please consult the [migration guide]({{< ref "user-guide/aws/lambda#migrating-to
...
-| `LAMBDA_DOCKER_FLAGS` | `-e KEY=VALUE`, `-v host:container`, `-p host:container`, `--add-host domain:ip` | Additional flags passed to Docker `run`\|`create` commands. Supports environment variables (also with `--env-file`, but the file has to be mounted into the LocalStack container), ports, volume mounts, extra hosts, networks, DNS servers, labels, ulimits, user, platform, and privileged mode. |
+| `LAMBDA_DOCKER_FLAGS` | `-e KEY=VALUE`, `-v host:container`, `-p host:container`, `--add-host domain:ip` | Additional flags passed to Docker `run`\|`create` commands. Supports environment variables (also with `--env-file`, but the file has to be mounted into the LocalStack container), ports, volume mounts, extra hosts, networks, DNS servers, labels, ulimits, user, platform, and privileged mode. The `--env-file` for Docker `run` and Docker Compose have different feature sets. The `--env-file` for Docker `run` is used directly, while `--compose-env-file` supports placeholders with `${}`, replacing quotes, etc. |
```
dfangl, line 210:
```suggestion
| `LAMBDA_DOCKER_FLAGS` | `-e KEY=VALUE`, `-v host:container`, `-p host:container`, `--add-host domain:ip` | Additional flags passed to Docker `run`\|`create` commands. Supports environment variables (also with `--env-file`, but the file has to be mounted into the LocalStack container), ports, volume mounts, extra hosts, networks, DNS servers, labels, ulimits, user, platform, and privileged mode. The `--env-file` argument for Docker `run` and Docker Compose have different feature sets. To provide both, we support the `--env-file` for environment files with the docker run syntax, while `--compose-env-file` supports the full docker compose features, like placeholders with `${}`, replacing quotes, etc. |
```

dfangl (approved):
Looks good, I just added a suggested change which makes it easier to understand in my opinion.

PR #1367, opened 2024-07-11
add set-token option to add the auth token

HarshCasper:


alexrashed (changes requested):
In general, I think it's a great idea to push the usage of the CLI and especially the `set-auth` command. However, it cannot really replace setting the environment variables when using other means to start LocalStack. I think it might be necessary to rewrite a bit more in the docs to make this distinction clear to the users, while using the full potential of the CLI?

content/en/user-guide/aws/dms/index.md
```diff
@@ -37,7 +37,7 @@ git clone https://github.com/localstack-samples/sample-dms-kinesis-rds-mariadb.g
 Next, start LocalStack (there is a docker-compose included, setting the `ENABLE_DMS=1` flag):
 
 ```sh
-export LOCALSTACK_AUTH_TOKEN=<your-auth-token> # this must be a enterprise license token
+localstack auth set-token <your-auth-token> # this must be a enterprise license token
```
alexrashed, line 40: This one still uses the auth token for `docker-compose` which won't work.

alexrashed (changes requested):

alexrashed (approved):
LGTM! This will make the `localstack auth set-token` command more prominent! 🚀

remotesynth (approved):
thanks for making this update!

PR #1368, opened 2024-07-11
add docs on EventBridge Scheduler

HarshCasper:


joe4dev (approved):
LGTM 👍 

minor suggestion: It's a bit awkward that the tagging operation doesn't return any response, which might be misleading (i.e., perceived as error). So either skip it (because tagging is completely generic and unrelated to the service) or consider showing the result via `awslocal scheduler list-tags-for-resource --resource-arn arn:aws:scheduler:us-east-1:000000000000:schedule/default/sqs-templated-schedule`

PR #1369, opened 2024-07-11
add docs on lakeformation

HarshCasper:


content/en/user-guide/aws/lakeformation/index.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Getting started
+
+This guide is designed for users new to Lake Formation and assumes basic knowledge of the AWS CLI and our [`awslocal`](https://github.com/localstack/awscli-local) wrapper script.
+
+Start your LocalStack container using your preferred method. We will demonstrate how to register an S3 bucket as a resource in Lake Formation, grant permissions to a user, and list the resources and permissions.
+
+### Register the resource
+
+Create a new S3 bucket named `lf-emr-athena-result-123` using the `mb` command:
```
silv-io, line 21: I think it might be easier to read if this was just a simple name. Otherwise people reading it might think there is some significance to this specific name.

content/en/user-guide/aws/lakeformation/index.md
```diff
@@ -0,0 +1,101 @@
...
+}
+```
+
+Run the following command to grant permissions:
+
+{{< command >}}
+$ awslocal lakeformation grant-permissions \
+    --cli-input-json file://check.json
+{{</ command >}}
+
```
silv-io, line 94: I don't know if that is the case here as well, but without setting `ENFORCE_IAM`, setting permissions is just "cosmetic". So if that is the case here as well, we should maybe hint at that and at the environment variable here.

silv-io (approved):
LGTM!

Just some minor comments

content/en/user-guide/aws/lakeformation/index.md
HarshCasper: Yeah, the user guides currently aren't at par with AWS since we skip mentioning the IAM enforcement feature or setting the right permissions. Will revisit this later.

PR #1373, opened 2024-07-12
document that the new event rule engine is default for new eventbridge provider

HarshCasper:


bartsdev:
You also need to make adjustment on this page: https://docs.localstack.cloud/user-guide/aws/pipes/

PR #1383, opened 2024-07-17
add basic docs on firelens

HarshCasper:


pinzon (approved):
Maybe we could link a sample of this feature if there is any.

simonrw (changes requested):
This feature is enterprise only, so we should mention this somewhere

marcciosilva (approved):

PR #1389, opened 2024-07-18
adapt information about default cloudwatch provider

steffyP:
Change note about the default provider.


pinzon (approved):
Looks good. We only need to fix the conflicts :+1:

PR #1392, opened 2024-07-22
fix parity update action

alexrashed:
## Motivation
We just recently adjusted some of our workflow names by introducing a prefix (`AWS / ...`).
This unfortunately broke the parity update workflow because it directly references the workflow by its name, [see this run](https://github.com/localstack/docs/actions/runs/10034961759).
This PR fixes this issue.

## Changes
- Adjust the workflow name for our pro parity metrics download.

### DIFF
--- a/.github/workflows/docs-parity-updates.yml
+++ b/.github/workflows/docs-parity-updates.yml
@@ -63,7 +63,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.PRO_ACCESS_TOKEN }}
           REPOSITORY_NAME: localstack-ext
           ARTIFACT_ID: parity-metric-ext-raw-*
-          WORKFLOW: "Build, Test, Push" 
+          WORKFLOW: "AWS / Build, Test, Push" 
           PREFIX_ARTIFACT: pro-integration-test
       
       - name: Download coverage (capture-notimplemented) data from Pro pipeline (GitHub)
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-parity-updates.yml
```diff
@@ -63,7 +63,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.PRO_ACCESS_TOKEN }}
           REPOSITORY_NAME: localstack-ext
           ARTIFACT_ID: parity-metric-ext-raw-*
-          WORKFLOW: "Build, Test, Push" 
+          WORKFLOW: "AWS / Build, Test, Push"
```
silv-io, line 66: I think to make this more resilient in the future, we might be able to refer to the workflow by its filename (`aws_main.yml`)

At least in the GitHub CLI you can choose the workflow by its name as well as its file name. 

In this case it still wouldn't help (since the file name also changed), but I feel that the file name would change less often than the workflow name.

silv-io (approved):
LGTM! Just a minor thought/suggestion.

.github/workflows/docs-parity-updates.yml
alexrashed: That's a good point, maybe in the future this would make sense, but as you mentioned right now, the filename is basically as stable as the name 😛

PR #1342, opened 2024-06-25
Fix parity update script

silv-io:
# Motivation

Since some changes in our core repositories happened, where the parity calculations don't happen on every commit, we're running into issues with the parity update script (see https://github.com/localstack/docs/actions/runs/9639842836). Now the parity reporting is always to be found in the latest scheduled run or manual full run of the pipelines.

# Changes

- Remove old handling of not found artifacts which relied on finding it on other executions on the same tree. The assumption that parity metrics are always in the latest commits or on the tree of dependent branches does not hold anymore. 
- Add loop based handling where the last 20 runs on `master` are looped through and searched for the artifact

# Testing

Running it locally now yields following result:

```
Searching for Artifact: parity-metric-ext-raw-* in workflow 'Build, Test, Push' on branch master in repo localstack/localstack-ext.
searching last workflow with 'conclusion=success'
Trying to download file with run_id 9659964244...
no artifact matches any of the names or patterns provided
Trying to download file with run_id 9656490933...
...downloaded parity-metric-ext-raw-* successfully.
./target/tmp_download
├── parity-metric-ext-raw-amd64-1
│   └── metric-report-raw-data-2024-06-25__04_49_41-pytest-junit-main-amd64-1.csv
├── parity-metric-ext-raw-amd64-2
│   └── metric-report-raw-data-2024-06-25__04_48_34-pytest-junit-main-amd64-2.csv
├── parity-metric-ext-raw-arm64-1
│   └── metric-report-raw-data-2024-06-25__04_50_08-pytest-junit-main-arm64-1.csv
└── parity-metric-ext-raw-arm64-2
    └── metric-report-raw-data-2024-06-25__04_49_27-pytest-junit-main-arm64-2.csv

5 directories, 4 files
Moving artifact to ./target/metrics-raw
content of ./target/metrics-raw:
./target/metrics-raw
├── pro-integration-test-metric-report-raw-data-2024-06-25__04_48_34-pytest-junit-main-amd64-2.csv
├── pro-integration-test-metric-report-raw-data-2024-06-25__04_49_27-pytest-junit-main-arm64-2.csv
├── pro-integration-test-metric-report-raw-data-2024-06-25__04_49_41-pytest-junit-main-amd64-1.csv
└── pro-integration-test-metric-report-raw-data-2024-06-25__04_50_08-pytest-junit-main-arm64-1.csv

1 directory, 4 files
Done.
```

We can see that the latest run does not contain the metrics but the one before that did.

[...]

PR #1415, opened 2024-08-05
fix parity upgrade action by updating CircleCI URL

alexrashed:
## Motivation
We recently had a few changes to our pipelines.
https://github.com/localstack/docs/pull/1342 already adjusted the Parity Update scripts to some of the changes.
Unfortunately, we've overlooked one change: The pipeline in CircleCI in `localstack/localstack` has been changed such that it is not called `main` anymore: https://github.com/localstack/localstack/pull/11049 renamed `main` to `full-run` and introduced a new `acceptance-only-run`.

Unfortunately, we haven't realized that this would have to change this here as well.
Now the artifacts of the old pipeline are not available anymore, which is why the run failed: https://github.com/localstack/docs/actions/runs/10242910192

## Changes
- This PR fixes the parity update action by referencing the correct workflow again.

## Testing
- [Triggered a run from this PR](https://github.com/localstack/docs/actions/runs/10248668104/job/28350431431) to verify that the download of the artifacts is fixed by the changes in this PR.
  - https://github.com/localstack/docs/pull/1416 was created by the run, which verifies that the fix is working.

### DIFF
--- a/scripts/get_latest_circleci_metrics.sh
+++ b/scripts/get_latest_circleci_metrics.sh
@@ -24,7 +24,7 @@ METRICS_IMPL="$PARENT_FOLDER/metrics-implementation-details/"
 
 echo "Project: $PROJECT_SLUG."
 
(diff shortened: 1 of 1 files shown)


steffyP (approved):
Thanks for jumping onto it 🙌

PR #1419, opened 2024-08-06
Add callouts for DNS configuration of TEI

simonrw:
Our documentation of Transparent Endpoint Injection does not cover:
* that it's implicitly disabled with the `DNS_ADDRESS=0` configuration, which is common in old examples, and still very present in users' configurations
* that it's required for e.g. CDK custom resources to function
    * actually if the lambda uses an SDK that supports `AWS_ENDPOINT_URL` then we don't technically need TEI, however
    * we also disable the required TLS verification when `DNS_ADDRESS=0'`.

[Rendered](https://localstack-docs-preview-pr-1419.surge.sh/user-guide/tools/transparent-endpoint-injection/)


dfangl (approved):
Good addition, this should clarify that functionality!

PR #1428, opened 2024-08-14
update ecr repo name on the eks docs

HarshCasper:


dfangl (commented):
There is a callout about the ECR port above the changes, could we adapt/remove this as well?

dfangl (approved):
LGTM, thanks for removing it so quickly!

bartsdev (approved):

PR #1432, opened 2024-08-16
add docs for dms serverless

steffyP:
adds a paragraph about dms serverless, and details for the configuration that can be set for serverless.

### DIFF
--- a/content/en/references/configuration.md
+++ b/content/en/references/configuration.md
@@ -105,6 +105,13 @@ This section covers configuration options that are specific to certain AWS servi
 | - | - | - |
 | `PROVIDER_OVERRIDE_CLOUDWATCH` | `v1` | Use the old CloudWatch provider. |
 
+### DMS
+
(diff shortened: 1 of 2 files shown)


sannya-singal (commented):
Thanks for adding serverless 🎉 One suggestion: we can add a section for Replication Configs, similar to the one we have for Replication Instances in the docs.

steffyP:
> One suggestion: we can add a section for Replication Configs, similar to the one we have for Replication Instances in the docs.

The section refers to the resource browser though, and we don't have a resource browser for serverless yet.

sannya-singal (approved):
Thanks for clarifying, maybe in that case we could create a backlog item for this and we are good to go for this PR then 👍

steffyP:
> Thanks for clarifying, maybe in that case we could create a backlog item for this and we are good to go for this PR then 👍

I have already created a backlog item last week 😄

PR #1436, opened 2024-08-19
 document sse-c encryption

HarshCasper:


bentsku (approved):
Thanks for adding this! 

We really need the feature catalog for S3, because we also support 2 other encryption types managed by AWS: `AES256` and `aws:kms`, which are incompatible with SSE-C (we should now almost support all encryption parameters for S3, basically). 

Thanks again!

PR #1438, opened 2024-08-19
Added credential provider example

siennathesane:
Added a credential provider example for the AWS for Go SDK v2. There is a matching PR with the complete code at localstack/localstack-aws-sdk-examples#9.

### DIFF
--- a/content/en/user-guide/integrations/sdks/go/index.md
+++ b/content/en/user-guide/integrations/sdks/go/index.md
@@ -78,6 +78,120 @@ func main() {
   })
   // ...
 }{{< /tab >}}
+
+{{< tab header="aws-go-sdk-v2 Credential Provider" lang="golang" >}}
+package main
+
+import (
+	"context"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+)
+
+const (
+	AwsLocalEndpoint        = "http://localhost:4566"
+	AwsLocalCredentialsName = "AwsLocalCredentials"
+	AwsLocalDefaultRegion   = "us-east-1"
+	AwsLocalAccountId       = "000000000000"
+	AwsLocalAccessKey       = "test"         
+	AwsLocalSecret          = "test"         
+)
+
+var (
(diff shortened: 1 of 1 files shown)


gregfurman:
Hey. Thanks for this! I'm always a fan of Golang + LocalStack getting more love 😄

Couple of remarks on this. First, the CI is failing becase of linting issues (see [failed job here](https://github.com/localstack/docs/actions/runs/10460856285/job/29042784318?pr=1438)). Also maybe give the [best practices](https://github.com/localstack/docs?tab=readme-ov-file#best-practices) a lookthrough which could help diagnose/fix this.

In addition, I think for the purposes of documentation, it would probably be a better idea to provide users' with a simpler minimum-viable-solution. While what this PR  proposes would obviously work, I'm concerned perhaps the scope is a bit large.

What about something like this?

```go
const (
	AwsLocalEndpoint        = "http://localhost:4566"
	AwsLocalCredentialsName = "AwsLocalCredentials"
	AwsLocalDefaultRegion   = "us-east-1"
	AwsLocalAccountId       = "000000000000"
	AwsLocalAccessKey       = "test"
	AwsLocalSecret          = "test"
)

func main() {
	// build an aws.Credential-compliant configuration
	localstackCredentials := aws.Credentials{
		AccessKeyID:     AwsLocalAccessKey,
		SecretAccessKey: AwsLocalSecret,
		SessionToken:    "SESSION",
		Source:          "LocalStack hardcoded credentials.",
	}

	awsCfg, err := config.LoadDefaultConfig(context.TODO(),
		config.WithCredentialsProvider(credentials.StaticCredentialsProvider{
			Value: localstackCredentials,
		}),
		config.WithDefaultRegion(AwsLocalDefaultRegion),
		// Optionally, add more LoadOptions...
	)

	// ...

}
```
It could probably be slimmed down even further to just focus on the credential provider.

What are your thoughts?

siennathesane:
The matching PR in the examples repo contains all the necessary code, so I don't think this update is required. I included it since it's a common use case, but I don't feel strongly about this PR. Feel free to update or close it, I'm fine with either

PR #1440, opened 2024-08-20
update APIGW docs on custom domain

bentsku:
This PR updates the documentation around API Gateway Custom Domains. The feature was not marked as Pro only. Also added a small paragraph showing the use of it.
\cc @thrau 

Page preview: https://localstack-docs-preview-pr-1440.surge.sh/user-guide/aws/apigateway/#custom-domain-names-with-api-gateway-pro

### DIFF
--- a/content/en/user-guide/aws/apigateway/index.md
+++ b/content/en/user-guide/aws/apigateway/index.md
@@ -367,14 +367,29 @@ Setting the API Gateway ID via `_custom_id_` works only on the creation of the r
 Ensure that you set the `_custom_id_` tag on creation of the resource.
 {{< /callout >}}
 
-## Custom Domain Names with API Gateway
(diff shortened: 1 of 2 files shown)


cloutierMat (commented):
We should also mark it as pro on the Feature Coverage page.

https://localstack-docs-preview-pr-1440.surge.sh/user-guide/aws/feature-coverage/

bentsku:
> We should also mark it as pro on the Feature Coverage page.

I've tried to update it, not very easy as those operations are supported in Community, just not "emulated", and it's not easy to make a difference on that page 😬

cloutierMat (approved):
Awesome! Thanks for addressing the comments. We can get to the full support of custom domains in the future.

HarshCasper (approved):

PR #1444, opened 2024-08-21
add docs for iot data

HarshCasper:


viren-nadkarni (changes requested):
Now that we have a dedicated page for this, we can move [this section](https://docs.localstack.cloud/user-guide/aws/iot/#device-shadows) here

viren-nadkarni (approved):

PR #1452, opened 2024-08-27
update the eksctl command

MarcelStranak:
For now, to ensure compatibility with the latest versions of `eksctl`, here is an update to the command to set the endpoints. This resolves the connectivity issues.


tinyg210 (approved):
nice! thanks for doing this, Marcel!

PR #1460, opened 2024-08-29
AppSync: document flag for specifying JS libs version

simonrw:
We introduced a flag for the user to control the version of the JS libs version to use with AppSync JavaScript resolvers. We never documented this feature beyond some support issues. This PR outlines how the flag works.


cloutierMat (approved):
:+1: Good catch! Nice to have it documented :pray:

PR #1467, opened 2024-09-03
Update index.md typo in Listener

drauedo:
It looks like there is a typo in the LB docs.


simonrw (approved):
That's great, thanks for fixing this!

PR #1470, opened 2024-09-05
add faq note about ssl issues due to revoked certificate

HarshCasper:


komarkovich (changes requested):
Thank you for drafting this @HarshCasper. But I would take a different approach here. This to me seems just as one of the updates we provided on the incident. We should only mention what needs to be done if they face the revoked local certificate for `localhost.localstack.cloud` issue. 
So I would get rid of the all the points except the first one:

> ### How do I resolve SSL issues due to revoked local certificate for `localhost.localstack.cloud`?
> To resolve the SSL issues due to revoked local certificate, we strongly recommend updating to the latest LocalStack version for the most reliable and seamless experience.

And add if after updating to the latest version they are still experiencing issues they should follow the steps: 

> 1. **Disable Certificate Download**: To prevent downloading a revoked certificate, set the environment variable `SKIP_SSL_CERT_DOWNLOAD=1`. This will cause LocalStack to use a self-signed SSL certificate. Additionally, it’s important to clear the cached certificate from your host machine. This can be done by deleting the cached certificate file. For example, on Linux systems, you can locate and remove the file at `~/.cache/localstack/volume/cache/server.test.pem`. The exact path may differ depending on your operating system and how you’ve started LocalStack. Please refer to our [documentation](https://docs.localstack.cloud/references/filesystem/#localstack-volume) for specific instructions.
> 2. **Use HTTP Instead of HTTPS**: Where possible, use `http://` instead of `https://` to avoid issues related to the revoked certificates. This workaround works with most browsers. However, Safari requires additional steps:
> 2.1. **Safari Users**: To make this work, you’ll need to first navigate to the page in a new tab and accept the security warning. To do this, make sure that LocalStack is started with `SKIP_SSL_CERT_DOWNLOAD=1` and that you have cleared the cached certificate as mentioned above. Once you’ve accepted the warning, you should be able to proceed.

We shouldn't mention those are temporary workarounds, but rather just steps to resolve the issue, as we discussed, we will implement a long-term fix for authenticated users, but that's not going to be the case for not-authenticated ones. 

That's all the info needed there, and adding all this additional information is just confusing as is not relevant at all (or won't be once the incident is resolved).

Would be happy to get @remotesynth input on this as well.

komarkovich (approved):
Looks better now 👍

PR #1472, opened 2024-09-05
Adding parameter substitution implementation example to Batch docs

drauedo:
Recently, we implemented parameter substitution to Batch. As [AWS](https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html#parameters) have its own way to do this, I believe the docs should illustrate how to configure parameter substitution to set default values.

### DIFF
--- a/content/en/user-guide/aws/batch/index.md
+++ b/content/en/user-guide/aws/batch/index.md
@@ -153,7 +153,7 @@ You should see the following output:
 }
 ```
 
-If you want to pass arguments to the command as [parameters](https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html#parameters), you can use the `Ref::` declaration to set placeholders for parameter substitution. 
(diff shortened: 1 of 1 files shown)


content/en/user-guide/aws/batch/index.md
```diff
@@ -153,6 +153,16 @@ You should see the following output:
 }
 ```
 
+In case that you would like to pass the arguments to the command as [`parameters`](https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html), it can be done by using the `Ref::` declaration to set placeholders for parameter substitution:
```
MarcelStranak, line 156:
```suggestion
If you want to pass arguments to the command as [parameters](https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html#parameters), you can use the `Ref::` declaration to set placeholders for parameter substitution. This allows the dynamic passing of values at runtime for specific job definitions.
```

MarcelStranak (approved):
LGTM

PR #1480, opened 2024-09-12
Update Crossplane docs to reflect latest CLI changes

finlaysawyer:
As per https://github.com/crossplane/crossplane/pull/4719, crossplane now has it's own standalone CLI and no longer ships `kubectl crossplane` via the `install.sh` script. This PR updates the docs to reflect this.


HarshCasper (approved):
Thank You 👍

PR #1497, opened 2024-09-24
Update command in ACM-PCA user-guide

abtreece:
Fixes a command that is attempting to append (>) to `tee` rather than pipe (|) to `tee`


viren-nadkarni (approved):
Thanks for fixing this @abtreece !

PR #1498, opened 2024-09-25
Fix hardcoded links and replace them with references

dfangl:
## Motivation

With recent PRs #1496 and #1490 we introduced some hard-coded links as bandages for link errors.

However, we should try to resort to ref shortcodes if possible.

## Changes
* Change links back to use the ref shortcode


drauedo (approved):
LGTM, thanks for fixing it.

PR #1515, opened 2024-10-08
Adding lenguage support note.

drauedo:
I believe that we should specify in our documentation which languages we provide support in. This will help to avoid future issues with non-supported languages.


MarcelStranak:
I'm not entirely convinced that this is necessary. I will leave this to you guys to decide.

HarshCasper (approved):

Anze1508:
Seems good

Anze1508 (commented):
seems good

komarkovich:
It doesn't hurt to mention it there, but I wouldn't add that as a callout, and maybe just mention it in a brief sentence "support is provided in English only" under the Support resources section.

PR #1516, opened 2024-10-09
add docs on supported regions

HarshCasper:


komarkovich (approved):
LGTM

content/en/references/regions-coverage.md
```diff
@@ -0,0 +1,54 @@
...
+description: >
+    This page lists the AWS Region Coverage for LocalStack's emulation of AWS services.
+cascade:
+  type: docs
+---
+
+## Introduction
+
+LocalStack ships with multi-region support, enabling users to emulate different AWS regions and namespace resources based on the region.
+With IAM Policy Enforcement, you can ensure that access control and permissions are applied accurately across supported regions.
```
dfangl, line 14: I am not sure about that statement, it does not seem to quite fit here? I mean yes, you can use it with IAM enforcement, but it should not make a difference.

dfangl (approved):
LGTM! Just one comment about mentioning IAM enforcement.

dfangl (commented):
I just double checked, the two region mentioned as not supported work fine for me (at least for sqs), could you verify that?

PR #1529, opened 2024-10-29
Update deploy script path in quickstart guide

tiurin:
Deploy script was moved in https://github.com/localstack-samples/sample-serverless-image-resizer-s3-lambda/pull/37


simonrw (approved):
Thank you for correcting this path!

PR #1531, opened 2024-10-30
Removing version attribute from docker compose config

drauedo:
The attribute `version` is obsolete and will be ignored when running the docker compose configuration.


MarcelStranak (approved):
LGTM! 🐳

PR #1542, opened 2024-11-05
Add note on debugging limitations for Lambda functions

MarcelStranak:
Attempt to improve the visibility on the remote debugging page.

### DIFF
--- a/content/en/user-guide/lambda-tools/debugging/index.md
+++ b/content/en/user-guide/lambda-tools/debugging/index.md
@@ -25,7 +25,7 @@ More examples and tooling support for local Lambda debugging (including support 
 
 {{< callout tip >}}
 Due to the ports published by the Lambda container for the debugger, it is currently only possible to debug one Lambda function at a time.
-For complex debugging scenarios that require multiple ports, please consult the [Lambda Debug Mode (preview)]({{< relref "debugging#lambda-debug-mode-preview" >}}) section.
+For advanced debugging scenarios, such as those requiring multiple ports, refer to [Lambda Debug Mode (preview)]({{< relref "debugging#lambda-debug-mode-preview" >}}) section.
 {{< /callout >}}
 
 ## Debugging Python lambdas
--- a/content/en/user-guide/lambda-tools/debugging/index.md
+++ b/content/en/user-guide/lambda-tools/debugging/index.md
@@ -23,6 +23,11 @@ More examples and tooling support for local Lambda debugging (including support 
 * [Lambda Debug Mode (preview)](#lambda-debug-mode-preview)
 * [Resources](#resources)
 
+{{< callout tip >}}
+Due to the ports published by the lambda container for the debugger, it is currently only possible to debug one Lambda function at a time.
+For complex debugging scenarios that require multiple ports, please consult the [Lambda Debug Mode (preview)]({{< relref "debugging#lambda-debug-mode-preview" >}}) section.
+{{< /callout >}}
+
 ## Debugging Python lambdas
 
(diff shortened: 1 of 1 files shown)


content/en/user-guide/lambda-tools/debugging/index.md
```diff
@@ -23,6 +23,11 @@ More examples and tooling support for local Lambda debugging (including support
 * [Lambda Debug Mode (preview)](#lambda-debug-mode-preview)
 * [Resources](#resources)
 
+{{< callout tip >}}
+Due to the ports published by the lambda container for the debugger, it is currently only possible to debug one Lambda function at a time.
```
joe4dev, line 27: nit: lambda -> Lambda

content/en/user-guide/lambda-tools/debugging/index.md
```diff
@@ -23,6 +23,11 @@ More examples and tooling support for local Lambda debugging (including support
...
+For complex debugging scenarios that require multiple ports, please consult the [Lambda Debug Mode (preview)]({{< relref "debugging#lambda-debug-mode-preview" >}}) section.
```
joe4dev, line 28: I wouldn't call it "complex debugging scenarios" because the limitation is pretty common (i.e., any scenario that creates more than one Lambda container).
Suggestion: Can we start with the problem (e.g., pointer to the error message) and then suggest the Lambda debug mode as a solution?

joe4dev (approved):
Great initiative for better visibility 🚀 

I added some clarification ideas

content/en/user-guide/lambda-tools/debugging/index.md
joe4dev: maybe not that import as the Lambda docs even call it [advanced configuration](https://docs.localstack.cloud/user-guide/lambda-tools/debugging/#advanced-configuration) 😅 

We miss instructions regarding the volume mount mapping there. I'll need to find some time to test all the samples and update these ...

content/en/user-guide/lambda-tools/debugging/index.md
MarcelStranak:
```suggestion
Due to the ports published by the Lambda container for the debugger, it is currently only possible to debug one Lambda function at a time.
```

content/en/user-guide/lambda-tools/debugging/index.md
MarcelStranak:
```suggestion
For advanced debugging scenarios, such as those requiring multiple ports, refer to [Lambda Debug Mode (preview)]({{< relref "debugging#lambda-debug-mode-preview" >}}) section.
```

MarcelStranak: >maybe not that import as the Lambda docs even call it [advanced configuration](https://docs.localstack.cloud/user-guide/lambda-tools/debugging/#advanced-configuration) 😅

Yes, that's why I chose the complex. I didn't believe it was only used for multi-port scenarios, but rather that it offers more possibilities for debugging and troubleshooting.

> Suggestion: Can we start with the problem (e.g., pointer to the error message) and then suggest the Lambda debug mode as a solution?

I can remove this line if it's too much.
```
Due to the ports published by the lambda container for the debugger, it is currently only possible to debug one Lambda function at a time.
```

joe4dev: Nice clarification 👍 

no need to remove the first line. I like that it explains the problem. I was just wondering whether we can make it more discoverable by mentioning the error message users get if they run into this issue.

All good, feel free to merge :)

PR #1544, opened 2024-11-06
add docs on localstack sdk for python

HarshCasper:


giograno (approved):
Looks good, thanks for contributing to this!
In the meanwhile, I also added support for SNS endpoints. I think we should look into ways to automatically create docs for the SDK (maybe look into something like Sphinx).

PR #1545, opened 2024-11-06
Remove callout about legacy CFn template deployer

simonrw:
We removed the legacy fallback of the template deployer with LocalStack version 3.6 so we should not mention this option in the docs.


dominikschubert (approved):
good catch!

PR #1546, opened 2024-11-07
Changing ip from ec2 doc's sample

drauedo:
With the current IP(172.17.0.X), the sample doesn't work. However, if we changed it to 127.0.01 it works as expected.

### DIFF
--- a/content/en/user-guide/aws/ec2/index.md
+++ b/content/en/user-guide/aws/ec2/index.md
@@ -172,7 +172,7 @@ Run the following command to test the Python Web Server:
 
 {{< command >}}
 $ curl 172.17.0.4:8000
-# or
+# Or, you can run:
 $ curl 127.0.0.1:29043
 {{< /command >}}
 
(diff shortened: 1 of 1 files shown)


viren-nadkarni:
Hi @drauedo this is incorrect because we want to always use the container IP from the Docker bridge. The proposed change will not work on for example Mac OS using Docker Desktop

content/en/user-guide/aws/ec2/index.md
```diff
@@ -171,7 +171,7 @@ You can now use the IP address to test the Python Web Server.
 Run the following command to test the Python Web Server:
 
 {{< command >}}
-$ curl 172.17.0.4:8000
+$ curl 127.0.0.1:8000
```
viren-nadkarni, line 174: What we could suggest is to use this alternative:

```suggestion
$ curl 127.0.0.1:8000
# or
$ curl 127.0.0.1:29043
```

HarshCasper (approved):

PR #1554, opened 2024-11-14
Update bedrock v4 docs

silv-io:
This PR updates the documentation to cover the new LLM-enabled functionality of Bedrock and the new configuration variables.

### DIFF
--- a/content/en/user-guide/aws/bedrock/index.md
+++ b/content/en/user-guide/aws/bedrock/index.md
@@ -15,17 +15,27 @@ The supported APIs are available on our [API Coverage Page](https://docs.localst
 
 This guide is designed for users new to AWS Bedrock and assumes basic knowledge of the AWS CLI and our `awslocal` wrapper script.
 
-Start your LocalStack container using your preferred method using the `LOCALSTACK_ENABLE_BEDROCK=1` configuration variable.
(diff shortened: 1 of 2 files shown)


content/en/user-guide/aws/bedrock/index.md
```diff
@@ -75,5 +86,5 @@ $ awslocal bedrock-runtime converse \
 
 ## Limitations
 
-* LocalStack Bedrock implementation is mock-only and does not run any LLM model locally.
+* LocalStack Bedrock currently only officially supports text-based models.
```
dfangl, line 89: What does "currently only officially" mean exactly?

dfangl (approved):
LGTM, thanks for the fast turnaround!

content/en/user-guide/aws/bedrock/index.md
silv-io: That at this point in time (=currently) we only officially support text-based models as opposed to image or other kinds of binary data models

dfangl: My question would be what kind of models are unofficially supported :sweat_smile: . It's fine for me now, this was just the first question popping in my head. For me they are either supported or not, but this phrasing leads me to believe other models might be unofficially supported :P

PR #1585, opened 2024-12-02
Fix typo in Initialization Hooks documentation

sparrowt:
`"state": "OK"` is not one of the documented states.

https://docs.localstack.cloud/references/init-hooks/#querying-stages
> A script can be in one of four states: UNKNOWN, RUNNING, SUCCESSFUL, ERROR. Scripts are by default in the UNKNOWN state once they have been discovered. The remaining states should be self-explanatory.

Fix the example response to be `"SUCCESSFUL"` which is what is actually returned.


HarshCasper (approved):
Thank You!

PR #1604, opened 2024-12-18
update APIGW URL contruction to use stageName

bentsku:
We've got a report by a user trying to access API Gateway `invoke-api` URL that it wouldn't work for them, because they constructed the API URL with Terraform, and used the `stage.id` instead of the `stage.name`. They directly reported looking at the documentation for this.

Updating this page to make this more correct.


cloutierMat (approved):
:rocket:

PR #1605, opened 2024-12-19
[ESM] Kafka does not yet support failure destinations

gregfurman:
We don't actually support failure destinations in Kafka as of yet. This PR changes the "implemented" to "not implemented".


joe4dev (approved):
Thanks for the clarification 👍

PR #1614, opened 2025-01-02
Added small clarification on profiles

anisaoshafi:
Added small clarification on configuration of profiles.
In my onboarding experience, the ./localstack/ directory was not created when I was trying profiles out and I spent some minutes getting lost on whether this directory was supposed to get automatically created or not. Which should be if you install localstack via cli.

### DIFF
--- a/content/en/references/configuration.md
+++ b/content/en/references/configuration.md
@@ -493,6 +493,7 @@ These configurations have already been removed and **won't have any effect** on 
 ## Profiles
 
 LocalStack supports configuration profiles which are stored in the `~/.localstack` config directory.
+If the directory does not exist, create it manually.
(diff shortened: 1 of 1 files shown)


anisaoshafi:
Seems the link checks are failing after today's upgrade of `DavidAnson/markdownlint-cli2-action@v18` --> `DavidAnson/markdownlint-cli2-action@v19` https://github.com/localstack/docs/commit/f319786719c3efa0d3a3a4fb9b934cf97dc28fa7 👀

bentsku:
\cc @HarshCasper any idea why the linting action now fails? Should we revert back to @18?

bentsku (approved):
LGTM, thanks for adding the clarification! I guess we can merge even with the lint failure.

PR #1622, opened 2025-01-07
Add s3 as ESM failure destination for Kinesis and DynamoDB stream sources

tiurin:
Implemented in https://github.com/localstack/localstack/pull/12087. This update to docs should be merged after ~~the next minor release~~ the next `latest` pro image is shipped.


joe4dev (approved):
> This update to docs should be merged after the next minor release once the feature is released.

@tiurin Feel free to merge this earlier after the next latest LocalStack Pro image is shipped 🚢 
Following our current practice, the docs reflect localstack@latest (we only use release branches for major releases so far).

PR #1631, opened 2025-01-22
add docs on rancher desktop

HarshCasper:


drauedo (approved):
LGTM 🚀 ! Thanks for having a look at my message yesterday and updating the docs accordingly. 🫶 🥇

bartsdev:
Some comments from me:
 - the placement of this instruction is not correct. It should be inside of the installation page.
 - I would divide the instructions based on OS (Mac/Linux, Windows) and simplify the guides
 - if you alias rancher to docker, shouldn't the localstack CLI work? If yes, then we should list that as an option.

bartsdev (changes requested):
see overal comment

PR #1650, opened 2025-02-07
Update index.md

hangonlyra:
Update Lambda troubleshooting to add workaround for Podman on MacOS


HarshCasper (changes requested):
Hi – Thanks for adding the patch! However, we have a dedicated Podman page here: https://docs.localstack.cloud/references/podman/

Is the information available on the page not enough to fix the issue? Let us know. In any case, we might need to move this information into the Podman docs to make sure that we have such workarounds documented.

PR #1579, opened 2024-11-24
Add a note about `NoEcho` parameter support in CloudFormation

rominf:
Depends on https://github.com/localstack/localstack/pull/11897.

PR #1653, opened 2025-02-13
CFn: document NoEcho behaviour

simonrw:
Duplicated from https://github.com/localstack/docs/pull/1579

We now have initial support for setting `NoEcho` in CFn parameters. This PR adds a callout reminding users not to use them for sensitive information, but otherwise they are availble.

Also: fix a linting error with the EC2 docs introduced in #1546


simonrw:
The previous PR was approved by me so I will merge

PR #1656, opened 2025-02-14
OpenSearch: Add missing config option

viren-nadkarni:
OpenSearch and ElasticSearch feature a config option to ignore failures during plugin initialisation. This config option was previously in the wrong section.

This PR moves the config option to the correct section to help discoverability. Furthermore it also adds its equivalent in ElasticSearch.


alexrashed (approved):
Thanks for adding the docs for these config options! 💯

PR #1665, opened 2025-02-24
add docs on shield

HarshCasper:


k-a-il (approved):
Thank you for adding docs for shield service!

PR #1678, opened 2025-03-03
reform the kubernetes docs

HarshCasper:
## Description

This PR:

- Removes the deprecated LocalStack on Kubernetes tutorial/repository
- Adds info on using Helm chart with LocalStack Pro and the supported parameters


alexrashed (changes requested):
Thanks a lot for removing the reference on the deprecated "LocalStack on K8s" repo! Pushing the helm chart is definitely the way to go! 🚀 
But I really want to avoid copying the current README from the helm chart to the docs here, as this is doomed to be outdated with the next release.

content/en/user-guide/integrations/kubernetes/index.md
```diff
@@ -54,98 +54,137 @@ Some useful Helm client commands are:
...
-            "id": "ses2pi5oap",
-            "name": "local-localstack-demo",
-...
+$ helm repo add localstack-charts https://localstack.github.io/helm-charts
+$ helm install my-release localstack-charts/localstack -f pro-values.yaml
 {{< /command >}}
 
-We can then use a browser to open the [Web UI](http://localhost:8081/archive-bucket/index.html), which should have been deployed to an S3 bucket inside LocalStack.
-The Web UI can be used to interact with the sample application, send new requests to the backend, inspect the state of existing requests, etc.
+## Parameters
```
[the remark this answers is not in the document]
HarshCasper: Automation has been added to pull the info directly from the Helm Chart's `README.md` file to this page.

layouts/shortcodes/github.html
alexrashed, line 1: **question:** The shortcodes / go templates are often really hard to read. The preview deployment is looking good, but this adds a lot of complexity to the existing shortcode.
Your new code seems to be very tailored to markdown and its section export. If I can see this correctly, this wouldn't work for any other language. I also think this isn't really compatible with the length parameter, would it?
In this case, I think it would make sense to just create a new one (maybe a `github-markdown` shortcode)?‏

content/en/user-guide/integrations/kubernetes/index.md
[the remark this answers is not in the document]
alexrashed: Pretty neat solution! It is worth mentioning, that changes to the helm chart would not directly trigger an update, but the docs are fairly often redeployed, and on every build the newest version of the chart's readme would be fetched! 🚀

alexrashed (commented):
If you decide that the README section really needs to be on the docs page, the usage of a shortcode is a great idea!
I think it might be better to create a new one though, I feel like the generic "add a set of line from github" doesn't necessarily have too much in common with the "add a specific section of a markdown file from GitHub" shortcode.

layouts/shortcodes/github.html
HarshCasper: Done!

HarshCasper: There have been some additional changes to the `github` shortcode since it was previously broken, and now works as expected!

alexrashed (approved):
Thanks for jumping on the comments and introducing this new shortcode to make sure that we fetch the data from GitHub at build-time! 🚀

PR #1687, opened 2025-03-07
Remove obsolete version top-level elements from compose examples

dfangl:
## Motivation
Reason for removing the version attributes see localstack/localstack#12354.

Also, hugo now takes the default docker compose files directly from the localstack/localstack repository using the github shortcode.

For this to work, I needed to enable the github shortcode to include the file as-is, without any highlighting, as the highlights are already done in the tab.
Happy to discuss this, if there is a better solution.

## Changes
* Remove `version` top level attributes from all docker compose examples.
* Get fresh versions of the docker compose examples directly from the LS repository.

### DIFF
--- a/content/en/getting-started/installation.md
+++ b/content/en/getting-started/installation.md
@@ -295,38 +295,10 @@ Docker Compose v1.9.0 and above is supported.
 
 {{< tabpane lang="yml" >}}
 {{< tab header="Community" lang="yml" >}}
-services:
-  localstack:
-    container_name: "${LOCALSTACK_DOCKER_NAME:-localstack-main}"
-    image: localstack/localstack
-    ports:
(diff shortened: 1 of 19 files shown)


dfangl:
Note that the preview still shows the version tag in the github-pulled files - I merged the PR after the build already started.

content/en/getting-started/installation.md
alexrashed, line 1: **praise:** Nice! This way they will be updated on every build! 🤩 ‏

alexrashed (approved):
Awesome! Thanks for the cleanup of the compose files! Getting rid of all the deprecation warnings! 🥳 💯

pinzon (approved):
Great!

HarshCasper (approved):

PR #1697, opened 2025-03-18
Update index.md

vijaynidhani:
Port info is missing, also it required environment variable change in LocalStack to listen on port 80 by setting the environment variable: GATEWAY_LISTEN=:4566,:80,:443

### DIFF
--- a/content/en/user-guide/aws/elb/index.md
+++ b/content/en/user-guide/aws/elb/index.md
@@ -84,6 +84,8 @@ The following command creates a listener for the load balancer created in the pr
 
 {{< command >}}
 $ listenerArn=$(awslocal elbv2 create-listener \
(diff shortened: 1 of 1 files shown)


nik-localstack (approved):
Thanks for your contribution @vijaynidhani 🚀 
Indeed these fields are now required after 4.2 version of localstack in parity to what is required in AWS

simonrw (commented):
You mention in the PR description that `GATEWAY_LISTEN` should be updated. Perhaps you can add a note to the elbv2 docs mentioning that if you want your load balancer to listen on port 80, then `GATEWAY_LISTEN` must include it?

This suggestion is not blocking to this PR, but I would like to see this added at some point cc @dfangl @nik-localstack

PR #1706, opened 2025-03-26
docs: fixed issue templates directory name and added config yml file

quetzalliwrites:
- Added a `config.yml` file to define the behavior of docs issue templates.
- Fixed issue template directory name.


.github/config.yml
```diff
@@ -0,0 +1,8 @@
+blank_issues_enabled: false
```
quetzalliwrites, line 1: `blank_issues_enabled: false` disables the "open a blank issue" button.

We need that because we want GitHub to show our new issue templates vs the default blank issue.

remotesynth (approved):

HarshCasper (approved):

PR #1718, opened 2025-04-01
Use get_latest_*_metrics.sh scripts from meta repository

dfangl:
## Motivation
We need the `get_latest_*_metrics.sh` scripts from different repositories now, so they have been added to the localstack/meta repository.

To avoid having them in multiple locations, we should get them from the meta repository in the docs workflows as well.

## Changes
* Use migrated `get_latest_*_metrics.sh` scripts from the meta repository


dfangl:
New run: https://github.com/localstack/docs/actions/runs/14193057557/job/39761931231 (cancelled it so it does not create a PR, but the download succeeded as expected)

alexrashed (approved):
Nice cleanup! Thanks for the push to make the scripts reusable and deliberately switching to its usage everywhere! 💯 🧹

PR #1734, opened 2025-04-16
Change message about supported action types of ELB

nik-localstack:
Simple change to specify the support of fixed-response action types


simonrw (approved):
Thanks for updating this

PR #1737, opened 2025-04-21
document limitations with the scheduler service

HarshCasper:
See https://github.com/localstack/localstack/issues/11263#issuecomment-2259090633


joe4dev (approved):
LGTM after fixing the linter errors

PR #1750, opened 2025-05-02
add a callout for new rds provider

HarshCasper:


nik-localstack (approved):
🚀

content/en/user-guide/aws/rds/index.md
```diff
@@ -16,6 +16,14 @@ RDS handles routine database tasks such as provisioning, patching, backup, recov
 LocalStack allows you to use the RDS APIs in your local environment to create and manage RDS clusters and instances for testing & integration purposes.
 The supported APIs are available on our [API coverage page](https://docs.localstack.cloud/references/coverage/coverage_rds/), which provides information on the extent of RDS's integration with LocalStack.
 
+{{< callout >}}
+We’ve introduced a new native RDS provider in LocalStack and made it the default.
+This replaces Moto-based CRUD operations with a more reliable setup.
+
+RDS state created in version 4.3 or earlier using Cloud Pods or standard persistence may not be compatible with the new provider introduced in version 4.4.
```
bentsku, line 23: I think it's almost 100% certain that the state won't be compatible, so I'm not too sure about "may not be compatible", we might want to be more assertive on the "breaking" part. Do you think that makes sense?

bentsku (approved):
LGTM! Just left a small comment about the "may break" part, as it most probably "will" break. But it's still good to go 😄

PR #1778, opened 2025-05-23
Clarify RDS postgres database name restriction

dfangl:
## Motivation
Before recent changes, using `postgres` as database name for an RDS postgres database lead to unexpected errors.

Now, it should work, but we would still like to encourage users to use a different database name.

## Changes
* Clarify comment about `postgres` database name


bentsku (commented):
LGTM! Thanks for following up! 🚀 leaving a docs owner give the final approval 😄

HarshCasper (approved):

PR #1790, opened 2025-06-12
S3 remote pods: add troubleshooting section

giograno:
This PR adds a troubleshooting section for S3 remotes.

It goes through a few issues recently mentioned by users (e.g., the `The AWS Access Key Id you provided does not exist in our records.` error).


komarkovich (approved):
LGTM

